PR #38, opened 2023-01-03
Implement wallet connect functionality

jubalm:
Wallet connect functionality
- [x] remove old implementation
- [x] move type definitions
- [x] create shared store for account states
- [x] connect button functionality
- [x] update utility styles

Resolves https://github.com/DarkFlorist/web3-app/issues/12
Resolves comments here https://github.com/DarkFlorist/web3-app/pull/16#pullrequestreview-1200452512

Suggested Changes:
- [x] generalized json rpc error handler
- [x] replaced address validations with ethers built-in
- [x] reimplement account change listener


app/ts/library/exceptions.ts
```diff
@@ -0,0 +1,23 @@
+export type MetaMaskRpcErrorSignature = {
```
KillariDev, line 1: This is called `EthereumJsonRpcError` in Interceptor. As it's not specific to Metamask only

app/ts/components/ConnectToWallet.tsx
```diff
@@ -0,0 +1,82 @@
...
+					</div>
+				</Wrapper>
+			)
+
+		case 'rejected':
+			if (account.error instanceof MetaMaskRpcError) {
+				return (
+					<Wrapper>
+						<div class='text-center md:text-right'>
+							<div class='font-bold'>Failed to connect to MetaMask!</div>
```
KillariDev, line 52:
```suggestion
							<div class='font-bold'>Failed to connect to a wallet!</div>
```

app/ts/components/ConnectToWallet.tsx
```diff
@@ -0,0 +1,82 @@
...
+							<div class='text-sm text-white/50'>
+								<a class='flex items-center gap-1 italic' title={`${account.error.message} (${account.error.code})`}>
+									<span>Open MetaMask extension window for details</span>️ <Icon.Info />
```
KillariDev, line 55:
```suggestion
									<span>Open your wallet extension window for details</span>️ <Icon.Info />
```

app/ts/store/account.ts
```diff
@@ -0,0 +1,79 @@
...
+
+const handleAccountChange = (newAccount: string[]) => {
+	if (store.value?.status !== 'connected') return
+	store.value = isAddress(newAccount[0]) ? { status: 'connected', address: newAccount[0] } : storeDefaults
+}
+
+// watch for account changes when an account is connected
+const observeAccountStatusChanges = () => {
+	if (!isEthereumObservable(window.ethereum)) return
+	window.ethereum.on('accountsChanged', handleAccountChange)
```
KillariDev, line 73: you should use `provider.on` here ethers method and not call it directly

https://docs.ethers.org/v5/api/providers/provider/#Provider--event-methods

app/ts/library/utilities.ts
```diff
@@ -20,3 +22,30 @@ export function removeNonStringsAndTrim(...strings: (string | boolean | undefine
...
 }
+
+/**
+ * Check string for valid address format
+ */
+export function assertsAddress(string: string): asserts string is `0x${string}` {
+	if (!/^0x[0-9A-Fa-f]{40}$/.test(string)) throw new Error(`Invalid address ${string}`)
+}
+
+export function isAddress(string: string): string is `0x${string}` {
```
KillariDev, line 33: ethers has `ethers.utils.isAddress()` which should be use instead

app/ts/library/utilities.ts
```diff
@@ -20,3 +22,30 @@ export function removeNonStringsAndTrim(...strings: (string | boolean | undefine
 			.join(' ')
 	)
 }
+
+/**
+ * Check string for valid address format
+ */
+export function assertsAddress(string: string): asserts string is `0x${string}` {
```
KillariDev, line 29: we can call `ethers.utils.isAddress` here

app/ts/library/utilities.ts
```diff
@@ -20,3 +22,30 @@ export function removeNonStringsAndTrim(...strings: (string | boolean | undefine
...
+
+/**
+ * Describe a window ethereum object
+ */
+type ExternalProvider = ethers.providers.ExternalProvider
+export function assertsExternalProvider(ethereum: unknown): asserts ethereum is ExternalProvider {
+	if (ethereum === null || ethereum === undefined || typeof ethereum !== 'object') throw new Error('Ethereum object does not exist')
+}
+
+export type EthereumWithListeners = {
```
KillariDev, line 45: no longer needed after we use ethers directly

app/ts/library/utilities.ts
```diff
@@ -20,3 +22,30 @@ export function removeNonStringsAndTrim(...strings: (string | boolean | undefine
...
+	on(eventName: string | symbol, listener: (...args: any[]) => void): void
+}
+
+export function isEthereumObservable(ethereum: unknown): ethereum is EthereumWithListeners {
```
KillariDev, line 49: same with this

KillariDev (approved):

app/ts/library/exceptions.ts
jubalm: The error I'm trying to capture actually only happens on MetaMask, thus I tried to capture the signature/shape to try and distinguish it. We may add a more general Ethereum Error capture later, at the moment only trying these out with MetaMask and Brave built-in wallets and figure a mechanism later when introducing broader support

app/ts/store/account.ts
jubalm: The `provider.on` seem to be used differently than intended here; to listen for `accountsChanged` event. There is an property `provider.provider` (subprovider in code) that basically exposes the configured provider (window.ethereum) Although it's type definitions do not have an `on` and `removeListener` so will still need narrowing to ensure they are methods that exist on the subprovider.

app/ts/library/utilities.ts
jubalm: I replaced this with `parseAddress`, which uses `ethers.utils.getAddress` as validator, as the `0x${string}` type needs to be satisfied for addresses

app/ts/library/utilities.ts
jubalm: this is still needed to narrow down ethers `provider.provider` to ensure the `on` method is present. `ExternalProvider` does not include one by default

app/ts/library/utilities.ts
jubalm: I removed `parseAddress` on the latest commit after changing the method used to get the active account to `signer.getAddress()`

app/ts/components/ConnectToWallet.tsx
```diff
@@ -0,0 +1,82 @@
...
+import * as Icon from './Icon/index.js'
+
+export const ConnectToWallet = () => {
+	const account = accountStore.value
+
+	switch (account.status) {
+		case 'disconnected':
+			return (
+				<Wrapper>
+					<div class='leading-tight text-right'>Quickly get started by connecting your wallet</div>
```
KillariDev, line 14:
```suggestion
					<div class='leading-tight text-right'>Get started quickly by connecting your wallet</div>
```

app/ts/components/ConnectToWallet.tsx
```diff
@@ -0,0 +1,82 @@
...
+					<Button class='border border-white/30 py-2 px-4 rounded leading-tight text-white/50'>Connecting...</Button>
+				</Wrapper>
+			)
+
+		case 'connected':
+			return (
+				<Wrapper>
+					<div class='flex gap-4 min-w-0'>
+						<div class='text-right min-w-0'>
+							<div class='text-sm text-white/50'>Your Wallet Address</div>
```
KillariDev, line 40:
```suggestion
							<div class='text-sm text-white/50'>Your Address</div>
```

KillariDev (approved):

Issue #40, opened 2023-01-04
Add functionality to trigger function on URL route change


jubalm:
Would be handy to fire a function on route change like reconnecting or checking connected account

Example:
```mermaid
sequenceDiagram
    User->>App: Navigates to /tx/send
    App-->>User: Show send form
    User->>App: Submits Form
    App-)User: Redirect to /tx/[TRANSACTION_HASH]
    note left of App: run function to reconnect account
    App-)Chain: Send Transaction
    Chain-->>App: Confirm Transaction
    App-->>User: Show transaction receipt
    
```

jubalm:
@MicahZoltu how do you feel about introducing [preact-router](https://www.npmjs.com/package/preact-router) instead of composing this ourselves? seems this has no dependencies

MicahZoltu:
> @MicahZoltu how do you feel about introducing [preact-router](https://www.npmjs.com/package/preact-router) instead of composing this ourselves? seems this has no dependencies

Does it support hash based routing?  For statically hosted SPAs, path based routing doesn't work because the browser will try to fetch the page at `/path/to/route` which won't exist since everything is just a dumb statically hosted SPA.

MicahZoltu:
Also, it is very cool that GitHub supports sequence diagrams!

MicahZoltu:
Why do you need to reconnect account when the route changes?

jubalm:
> Does it support hash based routing?

that will need history library then 🤔

> path based routing doesn't work because the browser will try to fetch the page at /path/to/route which won't exist since everything is just a dumb statically hosted SPA

i might be imagining things but doesn't IPFS has a redirect mechanism? that way setting `<base href='/' />` would correct the file references on a nested path?

> Why do you need to reconnect account when the route changes?

the dapp will reset it's states on reload/refresh. the dapp could then check if the wallet is unlocked and there's a connected account and reuse that instead of the user clicking connect again

jubalm:
Closing in favor of custom hash routing #43 and reconnect #51 strategy

PR #41, opened 2023-01-04
Replace routing with preact-router

jubalm:
Proof-of-concept to support route change triggers #40 by introducing preact-router library

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -1,11 +1,13 @@
 <!DOCTYPE html>
 
 <head>
+	<base href='/' />
 	<meta charset='utf-8'>
-	<link rel='stylesheet' href='./css/index.css'>
+	<meta name="viewport" content="width=device-width, initial-scale=1.0" />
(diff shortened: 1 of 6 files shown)


app/index.html
```diff
@@ -1,11 +1,13 @@
 <!DOCTYPE html>
 
 <head>
+	<base href='/' />
```
MicahZoltu, line 4: All paths should be relative as we don't know what the hosting environment will look like, and it may be hosted on a subpath.  On the same note, we can't setup a `<base href='/' />` because the file server may be serving on a subpath.

 (same comment for other paths in this PR)

MicahZoltu (commented):
Per conversation on today's call, we'll need to see if we can get purely hash based routing working as we want the app to work regardless of how/where it is hosted.  This means we can't do path based routing, and we also can't assume that the files will live at a particular path relative to the domain.

jubalm:
Closing this experiment in favor of Hash Routing #43

PR #42, opened 2023-01-07
Removal of es-module-shims dependency

jubalm:
Most modern browsers now support importmap except Safari. Open for discussion
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script/type/importmap

- [ ] added prettier config to prevent rewriting current code structure

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -18,7 +18,7 @@
 		}
 	</script>
 
-	<main></main>
+	<main>Loading...</main>
 
-	<script type='module' src='./js/index.js'></script>
+	<script async type='module' src='./js/index.js'></script>
 </body>
--- a/app/index.html
+++ b/app/index.html
@@ -1,15 +1,13 @@
 <!DOCTYPE html>
 
 <head>
-	<meta charset='utf-8'>
-	<link rel='stylesheet' href='./css/index.css'>
+	<meta charset="utf-8" />
+	<link rel="stylesheet" href="./css/index.css" />
 </head>
-<body style='margin: 0;'>
-	<script async type='module' src='./vendor/es-module-shims/es-module-shims.js'></script>
-	<script type='importmap'>
+<body>
+	<script type="importmap">
 		{
 			"imports": {
-				"es-module-shims": "./vendor/es-module-shims/es-module-shims.js",
 				"preact": "./vendor/preact/preact.module.js",
 				"preact/jsx-runtime": "./vendor/preact/jsx-runtime/jsxRuntime.module.js",
 				"preact/hooks": "./vendor/preact/hooks/hooks.module.js",
@@ -20,7 +18,7 @@
 		}
 	</script>
 
-	<main>Loading...</main>
+	<main></main>
 
-	<script type='module' src='./js/index.js'></script>
+	<script type="module" src="./js/index.js"></script>
 </body>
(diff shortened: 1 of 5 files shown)


app/index.html
```diff
@@ -1,15 +1,13 @@
 <!DOCTYPE html>
 
 <head>
-	<meta charset='utf-8'>
-	<link rel='stylesheet' href='./css/index.css'>
+	<meta charset="utf-8" />
+	<link rel="stylesheet" href="./css/index.css" />
```
MicahZoltu, line 5: Why did these get switched to double quotes?  Is the intent to change the style, or was this an accident?  IF it was intentional, it should be in a separate PR.

app/index.html
```diff
@@ -20,7 +18,7 @@
 		}
 	</script>
 
-	<main>Loading...</main>
+	<main></main>
```
MicahZoltu, line 21: Recommend retaining this `Loading...`.  In case the script loads slowly, this lets the user know that something is happening while they wait for Preact's first render.

build/vendor.ts
```diff
@@ -21,12 +23,14 @@ async function vendorDependencies() {
...
-	const importmapJson = JSON.stringify(importmap, undefined, '\t')
-		.replace(/^/mg, '\t\t')
+	const importmap = dependencyPaths.reduce(
+		(importmap, { packageName, entrypointFile }) => {
+			importmap.imports[packageName] = `./${path.join('.', 'vendor', packageName, entrypointFile).replace(/\\/g, '/')}`
+			return importmap
+		},
+		{ imports: {} as Record<string, string> }
+	)
+	const importmapJson = JSON.stringify(importmap, undefined, '\t').replace(/^/gm, '\t\t')
```
MicahZoltu, line 33: Why was this changed?  Removal of `es-module-shims` shouldn't have any impact on this file other than removing `es-module-shims` as a dependency.

app/index.html
```diff
@@ -20,7 +18,7 @@
...
 
-	<script type='module' src='./js/index.js'></script>
+	<script type="module" src="./js/index.js"></script>
```
MicahZoltu, line 23:
```suggestion
	<script async type="module" src="./js/index.js"></script>
```
While not required, we recently learned making this `async` is probably more correct and may help avoid some weird race conditions in FF.

MicahZoltu (changes requested):
There are some changes in this PR that seem unrelated to removing `es-module-shims` that should be either removed (if unintentional) or moved to a different PR.

app/index.html
jubalm: the editor automatically formats it to double quotes when saving even if prettier is already set to single quotes. amended

build/vendor.ts
jubalm: the editor formats them on save. reverting these lines to avoid confusion

app/index.html
MicahZoltu: Hmm, this suggests you have some editor setting configured.  I am able to save `.html` files in VSCode without it converting `'` to `"` for me.

app/index.html
jubalm: Was going to build a fancier one but yeah, will move it to another PR anyways so reverted for now

MicahZoltu (approved):

PR #43, opened 2023-01-10
Hash Routing

jubalm:
Build a url routing system based on location hash

- [x] create hash router component
- [x] support dynamic path (string starting with semicolons) `/#!/transaction/:dynamic_value`
- [x] capture dynamic path values <sup>[1]</sup>

<sup>[1]</sup> Exported a reusable function `useParams` to capture dynamic hash values `:dynamic_value`--word `\w+` preceeded by a semicolon

### DIFF
--- a/app/ts/components/App.tsx
+++ b/app/ts/components/App.tsx
@@ -3,8 +3,8 @@ import { Route, Router, useParams } from './HashRouter.js'
 export function App() {
 	return (
 		<Router>
-			<Route path='/#!/tx/send' component={Send} />
-			<Route path='/#!/tx/:transaction_hash' component={Transaction} />
+			<Route path='#send' component={Send} />
+			<Route path='#tx/:transaction_hash' component={Transaction} />
 			<Route path='/' component={Home} />
 		</Router>
 	)
@@ -14,7 +14,7 @@ const Home = () => {
 	return (
 		<div class='px-4'>
 			<h4 class='my-3'>Home Page</h4>
-			<a class='px-4 py-2 border' href='#!/tx/send'>
+			<a class='px-4 py-2 border' href='#send'>
 				Go to send page
 			</a>
 		</div>
@@ -25,7 +25,7 @@ const Send = () => {
 	return (
 		<div class='px-4'>
 			<h4 class='my-3'>Send Page</h4>
-			<a class='px-4 py-2 border' href='#!/tx/43ll'>
+			<a class='px-4 py-2 border' href='#tx/43ll'>
 				Simulate send transaction
 			</a>
 		</div>
--- a/app/ts/components/App.tsx
+++ b/app/ts/components/App.tsx
@@ -1,15 +1,10 @@
-import { Route, Router } from './HashRouter.js'
-import { SendEthPage } from './SendEthPage.js'
+import { Route, Router, useParams } from './HashRouter.js'
 
 export function App() {
-	function handleHashChange() {
-		// handle hash change here
-	}
-
 	return (
-		<Router onHashChange={handleHashChange}>
+		<Router>
 			<Route path='/#!/tx/send' component={Send} />
-			<Route path='/#!/tx/:transaction_hash' component={SendEthPage} />
+			<Route path='/#!/tx/:transaction_hash' component={Transaction} />
 			<Route path='/' component={Home} />
 		</Router>
 	)
@@ -17,18 +12,36 @@ export function App() {
 
 const Home = () => {
 	return (
-		<div>
-			<h4>Home</h4>
-			<a href='#!/tx/send'>send</a>
+		<div class='px-4'>
+			<h4 class='my-3'>Home Page</h4>
+			<a class='px-4 py-2 border' href='#!/tx/send'>
+				Go to send page
+			</a>
 		</div>
 	)
 }
 
 const Send = () => {
 	return (
-		<div>
-			<h4>Send</h4>
-			<a href='#!/tx/123123'>transaction</a>
+		<div class='px-4'>
+			<h4 class='my-3'>Send Page</h4>
+			<a class='px-4 py-2 border' href='#!/tx/43ll'>
+				Simulate send transaction
+			</a>
+		</div>
+	)
+}
+
+const Transaction = () => {
+	const params = useParams()
+	const hash = params && 'transaction_hash' in params && params.transaction_hash
+
+	return (
+		<div class='px-4'>
+			<h4 class='my-3'>Transaction Details</h4>
+			<div>
+				Transaction Hash: <strong>{hash}</strong>
+			</div>
 		</div>
 	)
 }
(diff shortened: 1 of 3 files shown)


MicahZoltu (commented):
Is this really all that is needed to do hash routing?  No libraries needed at all?

app/ts/components/HashRouter.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+export const Route = ({ component: Component }: RouteProps) => {
+	return <Component />
+}
+
+export const Router = ({ children, onHashChange }: RouterProps) => {
+	const [activeRoute, setActiveRoute] = useState(getActiveRoute())
+
+	function getActiveRoute() {
+		// ts does not assert correctly if these are combined into one function
```
jubalm, line 51: I might be missing something on typing array filtering, tho this works
```ts
const isVNode = (child: ComponentChild): child is VNode<any> => child instanceof Object
const isRouteComponent = (child: VNode<any>): child is VNode<RouteProps> => 'path' in child.props && 'component' in child.props
const childArray = toChildArray(children).filter(isVNode).filter(isRouteComponent)
// childArray is type VNode<RouteProps>

```
tho i expected this would also work
```ts
const isRouteComponent = (child: ComponentChild): child is VNode<RouterProps> => child instanceof Object && 'path' in child.props && 'component' in child.props
const childArray = toChildArray(children).filter(isRouteComponent)
// childArray is 'string' | 'number' | VNode<{}>
// which is from toChildArray

```

MicahZoltu: The root of the problem is:
```ts

Comment (jubalm) [reply app/ts/components/HashRouter.tsx]: opted in with custom typing `children` component as object array shying away from `toChildArray`

File: app/ts/components/App.tsx [hunk 2]
```diff
@@ -1,10 +1,47 @@
-import { Route, Router } from './Router.js'
-import { SendEthPage } from './SendEthPage.js'
+import { Route, Router, useParams } from './HashRouter.js'
 
 export function App() {
 	return (
 		<Router>
-			<Route path='/' component={SendEthPage} />
+			<Route path='/#!/tx/send' component={Send} />
+			<Route path='/#!/tx/:transaction_hash' component={Transaction} />
```
MicahZoltu, line 7: The plan is to drop the `!` right?  Can we also drop the extraneous `/` and do `#send` and `#tx/:transaction_hash` or something?  It feels very noisy having all of those slashes in there.

app/ts/components/HashRouter.tsx
```diff
@@ -0,0 +1,81 @@
...
+}
+
+function useHashChangeEffect(handler: (event: Event) => void) {
+	useEffect(() => {
+		window.addEventListener('hashchange', handler)
+		return () => window.removeEventListener('hashchange', handler)
+	}, [])
+}
+
+function isRouteComponent(child: object): child is VNode<Parameters<typeof Route>[0]> {
```
MicahZoltu, line 46: Thinking more on this, I think we should only assert that we have a `VNode` with a `path` property.  Are there some checks we can do to verify that we actually have a VNode?  Right now really all we are doing is asserting that we are returning something with a `path: string` property.
```suggestion
function isRouteComponent(child: object): child is VNode<{ path: string }> {
```

app/ts/components/App.tsx
jubalm: oh it already works with or without the hash bang as long as valid hash👌just updated the example

app/ts/components/HashRouter.tsx
jubalm: or would you suggest we just add the `component` check as well so it's exclusively for the `Route` component? the function name already suggests it should be

jubalm: > Are there some checks we can do to verify that we actually have a VNode
The only required prop for VNode is `children: ComponentChildren` which could also be another `VNode<any>`

MicahZoltu: I guess just add a check for `children` then.  😖

app/ts/components/App.tsx
```diff
@@ -1,10 +1,47 @@
-import { Route, Router } from './Router.js'
-import { SendEthPage } from './SendEthPage.js'
+import { Route, Router, useParams } from './HashRouter.js'
 
 export function App() {
 	return (
 		<Router>
-			<Route path='/' component={SendEthPage} />
+			<Route path='#send' component={Send} />
```
KillariDev, line 6:
```suggestion
			<Route path = '#send' component = { Send } />
```

did we disagree on styling here? 🤔

app/ts/components/HashRouter.tsx
```diff
@@ -0,0 +1,81 @@
...
+	const patternParts = stripLeadingSlash(pattern).split('/')
+	const pathParts = path.split('/')
+
+	if (patternParts.length !== pathParts.length) {
+		return null
+	}
+
+	return patternParts.reduce((params, part, index) => {
+		return part.startsWith(':') ? { ...params, [part.slice(1)]: pathParts[index] } : params
+	}, {} as { [key: string]: string })
```
KillariDev, line 80: can we get rid of `as` here?

KillariDev (approved):

app/ts/components/App.tsx
MicahZoltu: For TSX I prefer less spaces than in TS.

app/ts/components/HashRouter.tsx
MicahZoltu: It is necessary for `.reduce` to tell the type checker what type the `{}` is so it can correctly do type checking.

jubalm: reduce returns `{}` type without it

MicahZoltu, line 80: Consider: Simplify the function (drop curlies and return).
```suggestion
	return patternParts.reduce((params, part, index) => part.startsWith(':') ? { ...params, [part.slice(1)]: pathParts[index] } : params, {} as { [key: string]: string })
```
Alternatively, if performance is something we care about here:
```suggestion
	return patternParts.reduce((params, part, index) => {
		if (part.startsWith(':') params[part.slice(1)] = pathParts[index]
		return params
	}, {} as { [key: string]: string })
```

jubalm: good to know that spread operators are slower than regular assignment. the suggested snippet here has an error on 79? just missing closing pair. not sure if i can edit those before commit

app/ts/components/App.tsx
KillariDev: 😬

app/ts/components/HashRouter.tsx
```diff
@@ -0,0 +1,82 @@
...
+function matchPath(pattern: string, path: string): { [key: string]: string } | null {
+	const patternParts = stripLeadingSlash(pattern).split('/')
+	const pathParts = path.split('/')
+
+	if (patternParts.length !== pathParts.length) {
+		return null
+	}
+
+	return patternParts.reduce((params, part, index) => {
+		if (part.startsWith(':') params[part.slice(1)] = pathParts[index]
```
jubalm, line 79:
```suggestion
		if (part.startsWith(':')) params[part.slice(1)] = pathParts[index]
```

PR #45, opened 2023-01-17
Send Page UI Update

jubalm:
This PR sets the stage for the send page layout, separating the functionality to next PR

- Send Page UI layout and grid work
- fix type issue from #43 `Vnode<RouteProps>`

Sample UI screens

<small>mobile (portrait)</small>
mobile

<small>HD / tablets / mobile (landscape)</small>
tablet

<small>Full HD / laptops</small> 
desktop

<small>wide displays</small> 
ultrawide

### DIFF
--- a/app/ts/components/Blockie.tsx
+++ b/app/ts/components/Blockie.tsx
@@ -7,6 +7,9 @@ interface BlockieProps {
 	borderRadius?: string
 }
 
+const defaultSize = 8
+const defaultScale = 4
+
 function generateIdenticon(options: BlockieProps, canvasRef: HTMLCanvasElement) {
 	// NOTE -- Majority of this code is referenced from: https://github.com/alexvandesande/blockies
 	// Mostly to ensure congruence to Ethereum Mist's Identicons
@@ -103,17 +106,15 @@ function generateIdenticon(options: BlockieProps, canvasRef: HTMLCanvasElement) 
 		}
 	}
 
-	const opts = options || {}
-	const size = opts.size || 8
-	const scale = opts.scale || 4
-	const seed = opts.seed || Math.floor(Math.random() * Math.pow(10, 16)).toString(16)
+	const scale = options.scale || defaultScale
+	const seed = options.seed
 
 	seedrand(seed)
 
-	const color = opts.color || createColor()
-	const bgcolor = opts.bgColor || createColor()
-	const spotcolor = opts.spotColor || createColor()
-	const imageData = createImageData(size)
+	const color = createColor()
+	const bgcolor = createColor()
+	const spotcolor = createColor()
+	const imageData = createImageData(defaultSize)
 	const canvas = setCanvas(canvasRef, imageData, color, scale, bgcolor, spotcolor)
 
 	return canvas
@@ -126,7 +127,7 @@ export default function Blockie(props: BlockieProps) {
 			generateIdenticon(props, canvasRef.current)
 		}
 	}, [props.seed])
-	const dimension = (props.size || 8) * (props.scale || 4)
+	const dimension = defaultSize * (props.scale || 4)
 	return (
 		<canvas
 			ref={canvasRef}
--- a/app/ts/components/Blockie.tsx
+++ b/app/ts/components/Blockie.tsx
@@ -1,0 +1,144 @@
+// ported directly from https://github.com/DarkFlorist/TheInterceptor
+import { useRef, useEffect } from 'preact/hooks'
+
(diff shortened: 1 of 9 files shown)


app/ts/components/Blockie.tsx
```diff
@@ -0,0 +1,144 @@
...
+			if (imageData[i]) {
+				const row = Math.floor(i / width)
+				const col = i % width
+
+				cc!.fillRect(col * scale, row * scale, scale, scale)
+			}
+		}
+	}
+
+	const opts = options || {}
```
MicahZoltu, line 110: This parameter isn't optional, so there is no reason to `|| {}`, you can just use `options` directly below.

app/ts/components/Blockie.tsx
```diff
@@ -0,0 +1,144 @@
...
+	const size = opts.size || 8
+	const scale = opts.scale || 4
+	const seed = opts.seed || Math.floor(Math.random() * Math.pow(10, 16)).toString(16)
```
MicahZoltu, line 113: `seed` is a required parameter, so no need to `||` it with anything.
```suggestion
	const seed = opts.seed
```

app/ts/components/Blockie.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+	seedrand(seed)
+
+	const color = opts.color || createColor()
+	const bgcolor = opts.bgColor || createColor()
+	const spotcolor = opts.spotColor || createColor()
```
MicahZoltu, line 119: Do you know what these colors are for?  Blockie colors are strictly defined and shouldn't be overridden.  cc @KillariDev

app/ts/components/Blockie.tsx
```diff
@@ -0,0 +1,144 @@
...
+				const row = Math.floor(i / width)
+				const col = i % width
+
+				cc!.fillRect(col * scale, row * scale, scale, scale)
+			}
+		}
+	}
+
+	const opts = options || {}
+	const size = opts.size || 8
```
MicahZoltu, line 111: I think we want to remove `size` and only include `scale`.  I *believe* that changing the `size` will also change the resulting image which breaks the entire purpose of blockies.

Note: I know that you copied this code from elsewhere, but many of these options should not be configurable.

app/ts/components/Blockie.tsx
MicahZoltu: Dug into this.  All 3 of these should be non-overridable.  Recommend removing them from the `options` object entirely and just using `createColor()` to generate them.

app/ts/components/Blockie.tsx
```diff
@@ -0,0 +1,144 @@
...
+import { useRef, useEffect } from 'preact/hooks'
+
+interface BlockieProps {
+	seed: string
+	size?: number
+	scale?: number
+	color?: string
+	bgColor?: string
+	spotColor?: string
+	borderRadius?: string
```
MicahZoltu, line 11:
```suggestion
	seed: string
	scale?: number
	borderRadius?: string
```
Even `borderRadius` could probably be dropped and just leave it up to HTML to deal with rather than dealing with it in the blockie.

MicahZoltu (approved):

MicahZoltu:
Screenshot would be nice when doing UI work like this.  Sometimes my mental rendering system gets things a bit wrong.  😉

app/ts/components/Blockie.tsx
jubalm: removed logical `OR` and directly referenced `options`

app/ts/components/Blockie.tsx
jubalm: reworked to `options.seed` after update from previous comment

app/ts/components/Blockie.tsx
jubalm: > Dug into this. All 3 of these should be non-overridable. Recommend removing them from the `options` object entirely and just using `createColor()` to generate them.

applied on https://github.com/DarkFlorist/web3-app/pull/45/commits/20efdb65827be1b1f029c8aad7fc85c6431c0fba

app/ts/components/Blockie.tsx
jubalm: introduced `defaultSize` and `defaultScale` as constants instead of options, exposing `scale` for override on https://github.com/DarkFlorist/web3-app/pull/45/commits/20efdb65827be1b1f029c8aad7fc85c6431c0fba

PR #50, opened 2023-01-17
Workflow gh deploy

epheph:


jubalm (approved):

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,30 @@
...
+        with:
+          node-version: 'v18.13.0'
+          registry-url: 'https://registry.npmjs.org'
+      - name: Install and Build
+        run: |
+          npm run vendor
+          npm run build
+      - name: Deploy to GH Pages
+        run: |
+          git checkout -B gh-pages
```
jubalm, line 22: this will work on master branch?

epheph: Check here, it specifies what branch to operate on

https://github.com/DarkFlorist/web3-app/pull/50/files#diff-28802fbf11c83a2eee09623fb192785e7ca92a3f40602a517c011b947a1822d3R5

The `gh-pages` is just what github pages reads from and we push to that branch (via that checkout)

PR #51, opened 2023-01-17
Retain account session after page refresh

jubalm:
This should satisfy the requirement on #40

### DIFF
--- a/app/ts/store/account.ts
+++ b/app/ts/store/account.ts
@@ -26,7 +26,7 @@ type AccountConnectRejected = {
 export type AccountStore = AccountBusy | AccountConnected | AccountDisconnected | AccountConnectRejected
 
 const storeDefaults = { status: 'disconnected', connect, ensureConnected } as const
-const accountStore = signal<AccountStore>(storeDefaults)
+export const accountStore = signal<AccountStore>(storeDefaults)
 
 async function connect() {
 	try {
--- a/app/ts/store/account.ts
+++ b/app/ts/store/account.ts
@@ -15,6 +15,7 @@ type AccountConnected = {
 type AccountDisconnected = {
 	status: 'disconnected'
 	connect: () => void
+	ensureConnected: () => void
 }
 
 type AccountConnectRejected = {
@@ -24,7 +25,7 @@ type AccountConnectRejected = {
 
 export type AccountStore = AccountBusy | AccountConnected | AccountDisconnected | AccountConnectRejected
 
-const storeDefaults = { status: 'disconnected', connect } as const
+const storeDefaults = { status: 'disconnected', connect, ensureConnected } as const
 const store = signal<AccountStore>(storeDefaults)
 export const accountStore = store
 
@@ -56,6 +57,20 @@ async function connect() {
 		store.value = { status: 'rejected', error }
 	}
 }
+
+async function ensureConnected() {
+ 	try {
+ 		assertsExternalProvider(window.ethereum)
+ 		const provider = new ethers.providers.Web3Provider(window.ethereum)
+ 		const signer = provider.getSigner()
+ 		store.value = { status: 'busy' }
+ 		const address = await signer.getAddress()
+ 		store.value = { status: 'connected', address }
+ 	} catch (exception) {
+ 		store.value = { status: 'disconnected', connect, ensureConnected }
+ 	}
+ }
+
 
 const handleAccountChange = (newAccount: string[]) => {
 	if (store.value?.status !== 'connected') return
--- a/app/ts/store/account.ts
+++ b/app/ts/store/account.ts
@@ -14,8 +14,8 @@ type AccountConnected = {
 
 type AccountDisconnected = {
 	status: 'disconnected'
-	connect: () => void
-	ensureConnected: () => void
+	connect: () => Promise<void>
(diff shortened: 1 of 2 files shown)


app/ts/store/account.ts
```diff
@@ -15,6 +15,7 @@ type AccountConnected = {
 type AccountDisconnected = {
 	status: 'disconnected'
 	connect: () => void
+	ensureConnected: () => void
```
MicahZoltu, line 18: The implementation for both of these are async so the interface should be async as well.  The caller can decide not to await if they like, but they should have the option to.
```suggestion
	connect: () => Promise<void>
	ensureConnected: () => Promise<void>
```

app/ts/store/account.ts
```diff
@@ -57,6 +58,20 @@ async function connect() {
...
+ 	try {
+ 		assertsExternalProvider(window.ethereum)
+ 		const provider = new ethers.providers.Web3Provider(window.ethereum)
+ 		const signer = provider.getSigner()
+ 		store.value = { status: 'busy' }
+ 		const address = await signer.getAddress()
+ 		store.value = { status: 'connected', address }
+ 	} catch (exception) {
+ 		store.value = { status: 'disconnected', connect, ensureConnected }
+ 	}
```
MicahZoltu, line 71: Consider using `useAsyncState` rather than manually reconstructing the logic everywhere that you have some async work.  Same for `connect()` above.

MicahZoltu (changes requested):
It isn't clear to me how will this trigger on route change?  Dose `accountStore.value` in the `App()` function result in a re-render of `App` if `accountStore` changes?

jubalm:
> It isn't clear to me how will this trigger on route change? Dose `accountStore.value` in the `App()` function result in a re-render of `App` if `accountStore` changes?

you are not mistaken, this is more like "there isn't a necessity to trigger handlers every route change" anymore as the use-case i presented (which is the only scenario to fulfill at this point) is to keep the account connected. the necessity to reconnect on route change was becoming less useful as we are using hash routing anyways which preserves the listener/observer for account changes for the entirety of the browsing session.

`ensureConnected` is introduced as a "silent" version of connect to re-assign account if wallet is still connected or just reset account altogether.

app/ts/store/account.ts
```diff
@@ -57,6 +58,20 @@ async function connect() {
...
+async function ensureConnected() {
+ 	try {
+ 		assertsExternalProvider(window.ethereum)
+ 		const provider = new ethers.providers.Web3Provider(window.ethereum)
+ 		const signer = provider.getSigner()
+ 		store.value = { status: 'busy' }
+ 		const address = await signer.getAddress()
+ 		store.value = { status: 'connected', address }
+ 	} catch (exception) {
+ 		store.value = { status: 'disconnected', connect, ensureConnected }
```
jubalm, line 70: I just realized these are similar values

```suggestion
		store.value = storeDefaults
```

app/ts/store/account.ts
```diff
@@ -24,7 +25,7 @@ type AccountConnectRejected = {
 
 export type AccountStore = AccountBusy | AccountConnected | AccountDisconnected | AccountConnectRejected
 
-const storeDefaults = { status: 'disconnected', connect } as const
+const storeDefaults = { status: 'disconnected', connect, ensureConnected } as const
 const store = signal<AccountStore>(storeDefaults)
 export const accountStore = store
```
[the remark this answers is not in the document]
jubalm: should be applied on https://github.com/DarkFlorist/web3-app/pull/51/commits/aabf5dd3b347a170d321661dd4130a838f1a2474

app/ts/store/account.ts
```diff
@@ -53,11 +53,25 @@ async function connect() {
...
+async function ensureConnected() {
+ 	try {
+ 		assertsExternalProvider(window.ethereum)
+ 		const provider = new ethers.providers.Web3Provider(window.ethereum)
+ 		const signer = provider.getSigner()
+ 		accountStore.value = { status: 'busy' }
+ 		const address = await signer.getAddress()
+ 		accountStore.value = { status: 'connected', address }
+ 	} catch (exception) {
+ 		accountStore.value = { status: 'disconnected', connect, ensureConnected }
```
jubalm, line 69: I just realized this was already declared
```suggestion
		accountStore.value = storeDefaults
```

app/ts/store/account.ts
```diff
@@ -24,19 +25,18 @@ type AccountConnectRejected = {
 
 export type AccountStore = AccountBusy | AccountConnected | AccountDisconnected | AccountConnectRejected
 
-const storeDefaults = { status: 'disconnected', connect } as const
-const store = signal<AccountStore>(storeDefaults)
-export const accountStore = store
+const storeDefaults = { status: 'disconnected', connect, ensureConnected } as const
+const accountStore = signal<AccountStore>(storeDefaults)
```
jubalm, line 29: Missing export

```suggestion
export const accountStore = signal<AccountStore>(storeDefaults)
```

app/ts/store/account.ts
jubalm: Refactoring this seem to have some moving parts that needed to be tested outside of this changeset. created a separate issue #54

Issue #52, opened 2023-01-18
Needs a more functional splash screen message or icon


jubalm:
This could use design work and can be utilized to convey a message to the user while the app is initializing
<img width="467" alt="Screenshot 2023-01-18 at 10 30 57 AM" src="https://user-images.githubusercontent.com/1169838/213067391-b67e6eae-8fd5-4c7c-9c15-a34880742585.png">

jubalm:
- Add the Lunaria icon

PR #53, opened 2023-01-18
Prevent FOUC by adding splash screen

jubalm:
When the page is first loaded, a flash of un-styled content happens while resources are being downloaded. This PR aims to reduce that "flicker" or "jankiness" feel to the app. This effect will be more pronounced once the app introduces handlers for page load and refresh such as #51 which this PR aims to prevent.

Miscellaneous addition:
- used "Lunaria" as brand

NOT included on this PR:
- splash screen icon design #52 

Temporary design

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -2,9 +2,68 @@
 <html lang='en'>
 
 <head>
+	<title>Lunaria</title>
 	<meta charset='utf-8' />
 	<meta name='viewport' content='width=device-width, initial-scale=1.0' />
 	<link rel='stylesheet' href='./css/index.css' />
+	<style>
+		.splash-screen {
+			position: fixed;
+			top: 0;
+			right: 0;
+			bottom: 0;
+			left: 0;
+			background-color: black;
+			display: flex;
+			flex-direction: column;
+			align-items: center;
+			justify-content: center;
+			color: white;
+			visibility: visible;
+			opacity: 1;
+			transform: scale(1);
+			transition: all 400ms 400ms;
+			background: radial-gradient(145.6% 145.6% at 50% 0%, #33003c 0%, #10001f 55.73%, #000000 100%);
+		}
+
+		.splash-screen h1 {
+			margin-bottom: 0.5rem;
+			font-size: 1.5rem;
+			font-weight: bold;
+		}
+
+		.splash-screen--off {
+			visibility: hidden;
+			opacity: 0;
+			pointer-events: none;
+			transform: scale(1.5);
+		}
+
+		.splash-screen circle {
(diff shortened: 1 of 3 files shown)


app/index.html
```diff
@@ -21,7 +80,16 @@
...
+	<main></main>
+
+	<section class='splash-screen'>
+		<h1>Lunaria</h1>
+		<svg xmlns='http://www.w3.org/2000/svg' width='24' height='24' viewBox='0 0 24 24' fill='currentColor'>
+			<circle cx='4' cy='12' r='3' />
+			<circle cx='12' cy='12' r='3' />
+			<circle cx='20' cy='12' r='3' />
+		</svg>
+	</section>
```
MicahZoltu, line 92: If you put this in `main`, and have the fragment returned by the `App` component have a `main` for its root element, then preact will replace this with the app once React is ready to take over rendering.  This way you don't have to do the `useSplashEffect` thing to hide it.

app/ts/components/App.tsx
```diff
@@ -23,3 +26,10 @@ const Transaction = () => {
 		</div>
 	)
 }
+
+function useSplashEffect() {
+	useEffect(() => {
+		const element = document.querySelector('.splash-screen')
+		element?.classList.add('splash-screen--off')
+	}, [])
+}
```
MicahZoltu, line 35: Should this be renamed to `removeSplash()` or something, since I believe that is what it is trying to do?

MicahZoltu (approved):

app/index.html
jubalm: this was to add a subtle transition especially when refreshing the page which makes the page looks like it "flickers" and can be confusing for the user (they saw something happened but was too quick to visually read/analyze it). the effect should hide/reduce it by the fading out the splash screen positioned over the `main`.


splash-screen-2

app/ts/components/App.tsx
jubalm: this was indeed named before figuring the mechanism for animation. i'm fine renaming it tho i've seen common use of functions prefixed with `use` for hooks, if that's something that we should also adapt? some variations i can also think of:

- useSplashScreenExit
- useSplashScreenTransitionEffect
- useTransitionFromSplash
- useExitSplashScreenEffect

or any of which without the `use` prefix. does these seem fit to you?

app/index.html
MicahZoltu: Ah, OK.  My ability to read CSS is very poor, and I thought the transition was instant.

app/ts/components/App.tsx
jubalm: added a `classList` check before injecting the classname

Issue #54, opened 2023-01-18
Consider using `useAsyncState` for async work.


jubalm:
Consider using `useAsyncState` rather than manually reconstructing the logic everywhere that you have some async work.  Same for `connect()` above.

_Originally posted by @MicahZoltu in https://github.com/DarkFlorist/web3-app/pull/51#discussion_r1073036463_

jubalm:
Closing as resolved on #61

PR #57, opened 2023-01-26
Inspect Transaction and Search Page

jubalm:
WIP for Transaction details page

- [x] View details of a transaction
- [x] Automatically wait for confirmation to calculate transaction fee
- [x] Added a manual transaction search page
Screenshot 2023-01-26 at 10 53 06 AM

- [x] switch send eth page to new hook to utilize useAsyncState
- [x] handle UI for presenting errors to user

### DIFF
--- a/app/ts/components/ErrorBoundary.tsx
+++ b/app/ts/components/ErrorBoundary.tsx
@@ -1,26 +1,24 @@
-import { Component, JSX } from 'preact';
+import { Component, JSX, ErrorInfo } from 'preact'
 
 interface Props {
 	children: JSX.Element | JSX.Element[]
 }
 
-interface State {
-	hasError: boolean;
-	error: Error | null;
-}
+type State = { hasError: false } | { hasError: true; error: Error }
 
 class ErrorBoundary extends Component<Props, State> {
 	constructor(props: Props) {
-		super(props);
-		this.state = { hasError: false, error: null };
+		super(props)
+		this.state = { hasError: false }
 	}
 
 	static getDerivedStateFromError(error: Error) {
-		return { hasError: true, error };
+		return { hasError: true, error }
 	}
 
-	componentDidCatch(error: Error, errorInfo: object) {
-		console.error(error, errorInfo);
+	componentDidCatch(error: Error, errorInfo: ErrorInfo) {
+		// connect to logging service
+		console.error(error, errorInfo)
 	}
 
 	render() {
@@ -30,14 +28,14 @@ class ErrorBoundary extends Component<Props, State> {
 					<div class='border border-red-400/20 bg-red-400/5 p-4 text-center max-w-lg'>
 						<div class='text-xl font-bold'>Uh oh, this is unexpected...</div>
 						<div class='text-white/50 text-sm my-4'>Not to worry. Drop by our disord channel and provide us with the hint below so we can help resolve your issue faster. Appreciate your support and look forward to seeing you there.</div>
-						<div class='border border-red-400/20 px-4 py-3 bg-black/10 text-white/80 text-sm'>{this.state.error?.message}</div>
+						<div class='border border-red-400/20 px-4 py-3 bg-black/10 text-white/80 text-sm'>{this.state.error.message}</div>
 					</div>
 				</div>
-			);
+			)
 		}
 
-		return this.props.children;
+		return this.props.children
 	}
 }
 
-export default ErrorBoundary;
+export default ErrorBoundary
--- a/app/ts/components/ErrorBoundary.tsx
+++ b/app/ts/components/ErrorBoundary.tsx
@@ -1,0 +1,43 @@
+import { Component, JSX } from 'preact';
+
+interface Props {
+	children: JSX.Element | JSX.Element[]
+}
+
+interface State {
+	hasError: boolean;
+	error: Error | null;
+}
+
+class ErrorBoundary extends Component<Props, State> {
+	constructor(props: Props) {
+		super(props);
+		this.state = { hasError: false, error: null };
+	}
+
+	static getDerivedStateFromError(error: Error) {
+		return { hasError: true, error };
+	}
+
+	componentDidCatch(error: Error, errorInfo: object) {
+		console.error(error, errorInfo);
+	}
+
+	render() {
+		if (this.state.hasError) {
+			return (
+				<div class='w-full my-[10%] flex justify-center'>
+					<div class='border border-red-400/20 bg-red-400/5 p-4 text-center max-w-lg'>
+						<div class='text-xl font-bold'>Uh oh, this is unexpected...</div>
+						<div class='text-white/50 text-sm my-4'>Not to worry. Drop by our disord channel and provide us with the hint below so we can help resolve your issue faster. Appreciate your support and look forward to seeing you there.</div>
+						<div class='border border-red-400/20 px-4 py-3 bg-black/10 text-white/80 text-sm'>{this.state.error?.message}</div>
+					</div>
+				</div>
+			);
+		}
+
+		return this.props.children;
+	}
+}
+
+export default ErrorBoundary;
--- a/app/ts/components/ErrorBoundary.tsx
+++ b/app/ts/components/ErrorBoundary.tsx
@@ -22,19 +22,17 @@ class ErrorBoundary extends Component<Props, State> {
 	}
 
 	render() {
-		if (this.state.hasError) {
-			return (
-				<div class='w-full my-[10%] flex justify-center'>
-					<div class='border border-red-400/20 bg-red-400/5 p-4 text-center max-w-lg'>
-						<div class='text-xl font-bold'>Uh oh, this is unexpected...</div>
-						<div class='text-white/50 text-sm my-4'>Not to worry. Drop by our disord channel and provide us with the hint below so we can help resolve your issue faster. Appreciate your support and look forward to seeing you there.</div>
-						<div class='border border-red-400/20 px-4 py-3 bg-black/10 text-white/80 text-sm'>{this.state.error.message}</div>
-					</div>
-				</div>
-			)
-		}
+		if (!this.state.hasError) return this.props.children
 
-		return this.props.children
+		return (
+			<div class='w-full my-[10%] flex justify-center'>
+				<div class='border border-red-400/20 bg-red-400/5 p-4 text-center max-w-lg'>
+					<div class='text-xl font-bold'>Uh oh, this is unexpected...</div>
+					<div class='text-white/50 text-sm my-4'>Not to worry. Drop by our disord channel and provide us with the hint below so we can help resolve your issue faster. Appreciate your support and look forward to seeing you there.</div>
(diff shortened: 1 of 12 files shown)


jubalm:
still investigating on #58 (navigation issue) but this should still be testable on it's own

app/ts/library/utilities.ts
```diff
@@ -42,3 +42,17 @@ export function isEthereumObservable(ethereum: unknown): ethereum is EthereumWit
...
+
+export function assertsTransactionHash(hash: string): asserts hash is `0x${string}` {
+	if (!isTransactionHash(hash)) throw new Error('Invalid transaction hash')
+}
+
+export const calculateGasFee = (effectiveGasPrice: ethers.BigNumber, gasUsed: ethers.BigNumber) => {
+	const gasFeeBigNum = effectiveGasPrice.mul(gasUsed)
+	const gasFee = ethers.utils.formatEther(gasFeeBigNum)
+	return gasFee
+}
```
MicahZoltu, line 58: Not really an issue at the moment, but at some point you may want to split out the Ethereum specific utilities from the more generic utilities (like sleep, removeNonStringsAndTrim).

app/ts/store/transfer.ts
```diff
@@ -0,0 +1,70 @@
...
+	| {
+			status: 'signed'
+			transactionResponse: TransactionResponse
+			fetchTransactionReceipt: () => Promise<void>
+	  }
+	| {
+			status: 'confirmed'
+			transactionResponse: TransactionResponse
+			transactionReceipt: TransactionReceipt
+	  }
```
MicahZoltu, line 31: Similar question to elsewhere.  Why not `useAsyncState`?  Is this different in some significant way?

app/ts/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,43 @@
+import { Component, JSX } from 'preact';
```
KillariDev, line 1:
```suggestion
import { Component, JSX } from 'preact'
```

KillariDev: there's semi columns in many places, they could be removed

app/ts/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+interface Props {
+	children: JSX.Element | JSX.Element[]
+}
+
+interface State {
+	hasError: boolean;
+	error: Error | null;
```
KillariDev, line 9: is there a reason to use null instead of undefined?

app/ts/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,43 @@
...
+	static getDerivedStateFromError(error: Error) {
+		return { hasError: true, error };
+	}
+
+	componentDidCatch(error: Error, errorInfo: object) {
+		console.error(error, errorInfo);
+	}
+
+	render() {
+		if (this.state.hasError) {
```
KillariDev, line 27: use guard clause

app/ts/components/Layout.tsx
```diff
@@ -1,29 +1,101 @@
...
+const Branding = () => (
+	<div class='flex flex-col justify-center items-start grow'>
+		<div class='text-3xl font-bold'>Lunaria</div>
+		<div class='text-sm text-white/50'>Decentralized • Open • Non-custodial</div>
+	</div>
+)
+
+const SocialLinks = () => (
+	<div class='flex items-center gap-6 text-white/30'>
+		<a class='text-white/30 hover:text-white' href='/'>
```
KillariDev, line 61:
```suggestion
		<a class='text-white/30 hover:text-white' href='https://github.com/DarkFlorist'>
```

app/ts/components/Layout.tsx
```diff
@@ -1,29 +1,101 @@
...
+			Github
+		</a>
+		<a class='text-white/30 hover:text-white' href='/'>
```
KillariDev, line 64:
```suggestion
		<a class='text-white/30 hover:text-white' href='https://twitter.com/DarkFlorist'>
```

app/ts/components/TransactionDetailsPage.tsx
```diff
@@ -0,0 +1,221 @@
...
+			<Layout.Footer />
+		</Layout.Page>
+	)
+}
+
+const Main = () => {
+	const transfer = transferStore
+	const hash = extractTransactionHashFromParams()
+
+	if (!hash) return <TransactionDetailsInvalid />
```
KillariDev, line 27:
```suggestion
	if (hash === undefined) return <TransactionDetailsInvalid />
```
its dangerous to check value without comparison as many things are false in JS

app/ts/components/TransactionDetailsPage.tsx
```diff
@@ -0,0 +1,221 @@
...
+		resolveTransactionResponse(async () => {
+			if (!hash || transferStore.value.status !== 'idle') return
+			await transferStore.value.fetchTransactionByHash(hash)
+		})
+	}
+
+	switch (transactionResponse.state) {
+		case 'inactive':
+			fetchTransactionResponse()
+			return null
```
KillariDev, line 59:
```suggestion
			return undefined
```

app/ts/components/TransactionDetailsPage.tsx
```diff
@@ -0,0 +1,221 @@
...
+						</button>
+						<button class='hover:underline underline-offset-4 flex items-center gap-1' onClick={() => { }}>
+							<Icon.Copy />
+							<span>Copy Transaction Hash</span>
+						</button>
+					</div>
+				</div>
+			)
+		case 'resolved':
+			return null
```
KillariDev, line 104:
```suggestion
			return undefined
```

app/ts/components/TransactionDetailsPage.tsx
```diff
@@ -0,0 +1,221 @@
...
+			await transfer.fetchTransactionReceipt()
+		})
+	}
+
+	assertTransferStatus(transfer.status, 'signed')
+
+	switch (transactionReceipt.state) {
+		case 'inactive':
+			fetchTransactionReceipt()
+			return null
```
KillariDev, line 126:
```suggestion
			return undefined
```

app/ts/components/TransactionDetailsPage.tsx
```diff
@@ -0,0 +1,221 @@
...
+						</button>
+						<button class='hover:underline underline-offset-4 flex items-center gap-1' onClick={() => { }}>
+							<Icon.Copy />
+							<span>Copy Transaction Hash</span>
+						</button>
+					</div>
+				</div>
+			)
+		case 'resolved':
+			return null
```
KillariDev, line 170:
```suggestion
			return undefined
```

app/ts/components/TransactionPage.tsx
```diff
@@ -0,0 +1,57 @@
...
+			<Layout.Body>
+				<Main />
+			</Layout.Body>
+			<Layout.Footer />
+		</Layout.Page>
+	)
+}
+
+const Main = () => {
+	const hash = useSignal('')
```
KillariDev, line 20: magic value?

KillariDev (approved):

app/ts/store/transfer.ts
jubalm: this was the exploration of the use of signal with async stuff "globally" which on the discord discussion should be avoided. adding this to the issue list on #54

app/ts/components/ErrorBoundary.tsx
jubalm: Thinking about it, i found a clearer way to define this type, ensuring `error` can only be of `Error` type

```ts
type State = { hasError: false } | { hasError: true; error: Error }
```

FWIW tho, `null` was used in the last scenario because `error` would have been defined but has no value.

app/ts/components/ErrorBoundary.tsx
jubalm: can you elaborate on this? looks to me like a guard clause already or i may not be interpreting the terminology well enough

app/ts/components/TransactionDetailsPage.tsx
jubalm: just pushed commit using `null` instead because `extractTransactionHashFromParams` is actually `0x${string} | null`

app/ts/components/TransactionDetailsPage.tsx
jubalm: these returns (including other suggestions) are from a functional component was very common to use `null` when you don't want to display or render visually on page. Preact would handle these the same anyways (null, boolean, undefined), but be nice if we have a standard one across projects. Thoughts @MicahZoltu ?

app/ts/components/TransactionPage.tsx
jubalm: ambiguous variable name, it's a text input value starts with blank string. more verbose replacement to `hashFieldInput`

app/ts/components/ErrorBoundary.tsx
MicahZoltu: Guard clauses are inverted:
```ts
if (guardCondition) return ...
if (otherGuardCondition) throw ...
return <div> ... </div>
```
The idea is to minimize nesting by having guard clauses that quickly return at the start of the function, and then the rest of your function is "the happy path".  In this case, the change would be:
```ts
		if (!this.state.hasError) return this.props.children
		return (
			<div class='w-full my-[10%] flex justify-center'>
				<div class='border border-red-400/20 bg-red-400/5 p-4 text-center max-w-lg'>
					<div class='text-xl font-bold'>Uh oh, this is unexpected...</div>
					<div class='text-white/50 text-sm my-4'>Not to worry. Drop by our disord channel and provide us with the hint below so we can help resolve your issue faster. Appreciate your support and look forward to seeing you there.</div>
					<div class='border border-red-400/20 px-4 py-3 bg-black/10 text-white/80 text-sm'>{this.state.error?.message}</div>
				</div>
			</div>
		);
```

app/ts/components/TransactionDetailsPage.tsx
MicahZoltu: I have a preference for `undefined` everywhere, never `null`.  This is just because there are so few situations where there is a meaningful difference between the two, and I think there is value in having consistency.

app/ts/components/TransactionPage.tsx
MicahZoltu: I like `hashFieldInput` better. 😬

app/ts/components/TransactionDetailsPage.tsx
KillariDev: I only use nulls when I have some function returning null and I cannot change that :). Eg `browser.windows.create` can return null

jubalm: a function returning undefined isn't valid JSX tho. it only works when wrapped with a fragment `<>{undefined}</>`

Screenshot 2023-02-02 at 2 05 07 AM

MicahZoltu: Huh, I didn't know that `null` was valid JSX but `undefined` isn't!  I will need to think on this and potentially alter my position on `null`s in preact projects.  🤔

PR #59, opened 2023-01-31
ETH Balance Component

jubalm:
This will need to be applied after #57 as the layout may have been changed there.

Opened for review to enable notifications.

### DIFF
--- a/app/ts/components/EthBalance.tsx
+++ b/app/ts/components/EthBalance.tsx
@@ -1,27 +1,33 @@
-import { useCallback } from "preact/hooks";
-import { assertUnreachable } from "../library/utilities.js";
-import { accountStore } from "../store/account.js";
-import { useBalanceAsync } from "../store/balance.js";
-import * as Icon from "./Icon/index.js"
+import { useCallback } from 'preact/hooks'
+import { assertUnreachable } from '../library/utilities.js'
+import { accountStore } from '../store/account.js'
+import { useBalanceAsync } from '../store/balance.js'
+import * as Icon from './Icon/index.js'
 
 export const EthBalance = () => {
-	const balance = useBalanceAsync()
+	const ethBalance = useBalanceAsync()
 
 	const checkBalance = useCallback(() => {
-		if (accountStore.value.status !== 'connected' || balance.state !== 'inactive') return;
-		balance.checkEthBalance()
+		if (accountStore.value.status !== 'connected' || ethBalance.state !== 'inactive') return
+		ethBalance.checkEthBalance()
 	}, [accountStore.value.status])
 
-	if (accountStore.value.status === 'disconnected') return (
-		<div class='flex flex-col items-center justify-center md:items-start md:gap-1 mb-2'>
-			<div class='uppercase text-white/50 text-xs md:text-sm flex'>balance (ETH)</div>
-			<div class='flex items-center gap-2'>
-				<div><button class="font-bold text-white/80 hover:text-white" onClick={accountStore.value.connect}>Connect wallet</button> <span class="text-white/50">to view balance</span></div>
+	if (accountStore.value.status === 'disconnected')
+		return (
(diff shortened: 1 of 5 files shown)


app/ts/store/balance.ts
```diff
@@ -0,0 +1,48 @@
...
+			state: 'rejected'
+			error: Error
+			reset: () => void
+	  }
+	| {
+			state: 'resolved'
+			balance: string
+			lastChecked: Date
+			reset: () => void
+	  }
```
MicahZoltu, line 25: This code *looks* a lot like `useAsyncState`.  I'm curious what the difference is, and why you chose to roll this instead of using `useAsyncState`?

app/ts/components/EthBalance.tsx
```diff
@@ -0,0 +1,58 @@
+import { useCallback } from "preact/hooks";
+import { assertUnreachable } from "../library/utilities.js";
+import { accountStore } from "../store/account.js";
+import { useBalanceAsync } from "../store/balance.js";
+import * as Icon from "./Icon/index.js"
```
KillariDev, line 5:
```suggestion
import { useCallback } from 'preact/hooks'
import { assertUnreachable } from '../library/utilities.js'
import { accountStore } from '../store/account.js'
import { useBalanceAsync } from '../store/balance.js'
import * as Icon from './Icon/index.js'
```

app/ts/components/EthBalance.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+export const EthBalance = () => {
+	const balance = useBalanceAsync()
+
+	const checkBalance = useCallback(() => {
+		if (accountStore.value.status !== 'connected' || balance.state !== 'inactive') return;
```
KillariDev, line 11:
```suggestion
		if (accountStore.value.status !== 'connected' || balance.state !== 'inactive') return
```

app/ts/components/EthBalance.tsx
```diff
@@ -0,0 +1,58 @@
...
+			<div class='flex items-center gap-2'>
+				<div><button class="font-bold text-white/80 hover:text-white" onClick={accountStore.value.connect}>Connect wallet</button> <span class="text-white/50">to view balance</span></div>
+			</div>
+		</div>
+	)
+
+	switch (balance.state) {
+		case 'inactive':
+			checkBalance()
+			return null
```
KillariDev, line 27:
```suggestion
			return undefined
```

app/ts/store/balance.ts
```diff
@@ -0,0 +1,48 @@
...
+
+async function checkEthBalance() {
+	try {
+		assertsExternalProvider(window.ethereum)
+		const provider = new ethers.providers.Web3Provider(window.ethereum)
+		if (accountStore.value.status !== 'connected') throw new Error('Not connected!')
+		balanceStore.value = { state: 'pending', reset }
+		const balanceBigNumber = await provider.getBalance(accountStore.value.address)
+		const balance = ethers.utils.formatEther(balanceBigNumber)
+		balanceStore.value = { state: 'resolved', balance, lastChecked: new Date(), reset }
```
KillariDev, line 35: you should store ether numbers as a big int and only when doing rendering, show it as a string

KillariDev (approved):

app/ts/store/balance.ts
jubalm: #54 will be working on this soon

app/ts/components/EthBalance.tsx
jubalm: will be refactoring on #63

app/ts/store/balance.ts
jubalm: This will be applied on #70

jubalm:
Closing this PR in favor of #70 due to tightly coupled dependency of the now updated signals-designed account state management.

PR #61, opened 2023-02-01
New state handling using useAsyncState & signals

### LINKED ISSUE #58: Fix render issue when using back and forward button
Investigate why the back and forward buttons do not seem to re-render the the send page and transaction page accordingly

jubalm:
Update `useAsyncState` with signals and applied to all stores

Resolves #58


app/ts/components/SendEthPage.tsx
```diff
@@ -48,40 +44,61 @@ const Main = () => {
...
+
+		if (!accountStore.value.isConnected) {
+			accountStore.value.connectMutation.dispatch()
+			return
+		}
+
+		if (!transfer.value.isSigned) {
+			transfer.value.transactionResponseQuery.dispatch()
+			return
+		}
```
MicahZoltu, line 70: It feels like we should tell the user about this rather than silently failing to submit the transaction.

app/ts/components/SendEthPage.tsx
```diff
@@ -48,40 +44,61 @@ const Main = () => {
...
-			)
-		default: assertUnreachable(transactionResponse)
+			return null
+		default:
+			assertUnreachable(transaction.value)
 	}
 }
 
-const Guide = ({ title, quote, content }: { title: string; quote?: string, content: string }) => {
+const Guide = ({ title, quote, content }: { title: string; quote?: string; content: string }) => {
```
MicahZoltu, line 101: Style: Consistency.  I think we use commas everywhere else?
```suggestion
const Guide = ({ title, quote, content }: { title: string, quote?: string, content: string }) => {
```

app/ts/store/account.ts
```diff
@@ -1,77 +1,89 @@
...
+		dispatch: () => {
+			waitFor(async () => {
+				assertsExternalProvider(window.ethereum)
+				const provider = new ethers.providers.Web3Provider(window.ethereum)
+				await provider.send('eth_requestAccounts', [])
+				const signer = provider.getSigner()
+				const address = await signer.getAddress()
+				accountStore.value = { isConnected: true, address }
+				return address
+			})
```
MicahZoltu, line 63: Consider removing the excess curlies here.
```suggestion
		dispatch: () => waitFor(async () => {
			assertsExternalProvider(window.ethereum)
			const provider = new ethers.providers.Web3Provider(window.ethereum)
			await provider.send('eth_requestAccounts', [])
			const signer = provider.getSigner()
			const address = await signer.getAddress()
			accountStore.value = { isConnected: true, address }
			return address
		)
```

app/ts/store/account.ts
```diff
@@ -1,77 +1,89 @@
...
 
-const storeDefaults = { status: 'disconnected', connect, ensureConnected } as const
-export const accountStore = signal<AccountStore>(storeDefaults)
+		reset()
+		accountStore.value = accountDefaults
+	}
 
-async function connect() {
-	try {
+	effect(() => {
```
MicahZoltu, line 81: Why is this inside an `effect` rather just inline?  Do you perhaps want `effect(..., [])` so it only runs once per re-render of anything that uses it?

MicahZoltu (approved):

app/ts/components/SendEthPage.tsx
```diff
@@ -48,40 +44,61 @@ const Main = () => {
...
+	return <form onSubmit={handleSubmit}>{children}</form>
+}
+
 const SendGuide = () => {
-	const { asyncResponse } = useTransfer()
-	const [transactionResponse] = asyncResponse
+	const transferStore = useTransferStore()
 
-	switch (transactionResponse.state) {
+	if (transferStore.value.isSigned) return <Guide title='Request Successfully Sent!' content="Your transaction is awaiting confirmation from the chain. You may click on the View Transaction button to check it's status." />
```
KillariDev, line 82: I believe we should be updating the transaction status all the time without user input?

app/ts/store/account.ts
```diff
@@ -1,77 +1,89 @@
...
+		dispatch: () => {
+			reconnect.waitFor(async () => {
+				try {
+					assertsExternalProvider(window.ethereum)
+					const provider = new ethers.providers.Web3Provider(window.ethereum)
+					const signer = provider.getSigner()
+					const address = await signer.getAddress()
+					accountStore.value = { isConnected: true, address }
+				} catch {
+					accountStore.value = accountDefaults
```
KillariDev, line 45: maybe print the error here and then handle it better later? This kind of all catchers can cause debugging to be hard later

KillariDev (approved):

app/ts/store/account.ts
```diff
@@ -1,77 +1,89 @@
...
+type Account =
+	| {
+			isConnected: false
+			connectMutation: ConnectMutation
+			reconnectMutation: ReconnectMutation
+	  }
+	| {
+			isConnected: true
+			address: string
+	  }
```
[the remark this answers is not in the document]
jubalm: this will require more than just changes from this file. created a separate PR for this suggestion #64

app/ts/components/SendEthPage.tsx
jubalm: Since this is using the account store, Components using it responds to the error; submit button text says "unable to connect" and the header button shows an error. I am inclined to this as well there should be a better way to announce this to the user. created a separate issue a #65

app/ts/components/SendEthPage.tsx
jubalm: Somewhere else means another project? Auto-format transforms it to semi-colons just as I changed it so I doubt it's in this project. Created a separate PR if we are going to use a semicolon convention on types #66

app/ts/store/account.ts
jubalm: I'm not aware `effect` accepting a second parameter looking at it's type, tho you are correct in assuming i'd only need to fire this once upon mount, so I suppose preact's `useEffect` would suffice.

app/ts/components/SendEthPage.tsx
jubalm: There is a separate page to view and track transaction status, would you suggest we automatically direct the user to that page?

app/ts/store/account.ts
jubalm: created a separate issue here #67

app/ts/store/account.ts
MicahZoltu: Yeah, I was thinking of `useEffect`, which would limit the frequency this is called I believe.

app/ts/components/SendEthPage.tsx
KillariDev: I mean instead of asking user to click View to see the status, you could show the status directly

app/ts/store/account.ts
```diff
@@ -1,77 +1,88 @@
...
+		dispatch: () => waitFor(async () => {
+			assertsExternalProvider(window.ethereum)
+			const provider = new ethers.providers.Web3Provider(window.ethereum)
+			await provider.send('eth_requestAccounts', [])
+			const signer = provider.getSigner()
+			const address = await signer.getAddress()
+			accountStore.value = { isConnected: true, address }
+			return address
+		)
+		},
```
0xjimmy, line 63:
```suggestion
		}),
```
Brace and parentheses in wrong order causing build to fail.

app/ts/components/SendEthPage.tsx
jubalm: created a separate feature issue as i have some additions to add and test to the transitioned view. #68

Issue #65, opened 2023-02-08
Better handling for Send Form's failed connection


jubalm:
this will require more than just changes from this file. created a separate PR for this suggestion #64

_Originally posted by @jubalm in https://github.com/DarkFlorist/web3-app/pull/61#discussion_r1099520247_

Issue #64, opened 2023-02-08
Separate account into 4 states


jubalm:
It feels like this should be more than just two states (connected/not-connected).  I can think of at least 4 possible states in reality:
1. Connect.
2. Not connected.
3. Connecting.
4. Connection failed.

You sort of have this indirectly via `*mutation.transport`, but that makes it difficult for users of the account to quickly and easily identify what the actual state is, especially since there is some weird overlaps between connect and reconnect.
```suggestion
type UnconnectedAccount = { state: 'unconnected', connect: () => void }
type ConnectingAccount = { state: 'connecting', reset: () => void }
type ConnectedAccount = { state: 'connected', address: string }
type ErroredAccount = { state: 'failed', error: Error, connect: () => void }
type Account = ConnectedAccount | UnconnectedAccount | ConnectingAccount | ErroredAccount
```

_Originally posted by @MicahZoltu in https://github.com/DarkFlorist/web3-app/pull/61#discussion_r1098192103_

jubalm:
Closing with #61 merged

PR #69, opened 2023-02-09
Account store state rework

### LINKED ISSUE #67: Decouple reconnect error handling
maybe print the error here and then handle it better later? This kind of all catchers can cause debugging to be hard later

_Originally posted by @KillariDev in https://github.com/DarkFlorist/web3-app/pull/61#discussion_r1098390971_

jubalm:
Addresses #64 which segregate address states

This also resolves #67 #65

### DIFF
--- a/app/ts/components/ConnectToWallet.tsx
+++ b/app/ts/components/ConnectToWallet.tsx
@@ -31,7 +31,7 @@ export const ConnectToWallet = () => {
 						<div class='transition animate-bounce-x'>
 							<Icon.ArrowRight />
 						</div>
-						<Button class='whitespace-nowrap' onClick={accountStore.value.connect.bind(null, false)}>
+						<Button class='whitespace-nowrap' onClick={accountStore.value.connect.bind(undefined, false)}>
 							Connect Wallet
 						</Button>
 					</div>
--- a/app/ts/components/ConnectToWallet.tsx
+++ b/app/ts/components/ConnectToWallet.tsx
@@ -24,7 +24,6 @@ export const ConnectToWallet = () => {
 				</Wrapper>
 			)
 		case 'disconnected': {
-			const connectToWallet = accountStore.value.connect
 			return (
 				<Wrapper>
 					<div class='grid grid-cols-[minmax(auto,max-content)_minmax(auto,max-content)_minmax(auto,max-content)] gap-3 items-center justify-center'>
@@ -32,7 +31,7 @@ export const ConnectToWallet = () => {
 						<div class='transition animate-bounce-x'>
 							<Icon.ArrowRight />
 						</div>
-						<Button class='whitespace-nowrap' onClick={() => connectToWallet()}>
+						<Button class='whitespace-nowrap' onClick={accountStore.value.connect.bind(null, false)}>
 							Connect Wallet
 						</Button>
 					</div>
--- a/app/ts/components/ConnectToWallet.tsx
+++ b/app/ts/components/ConnectToWallet.tsx
@@ -1,7 +1,5 @@
 import { ComponentChildren } from 'preact'
 import { EthereumJsonRpcError } from '../library/exceptions.js'
-import { assertUnreachable } from '../library/utilities.js'
-import { ConnectMutation } from '../store/account.js'
 import { useAccountStore } from './AccountContext.js'
 import Blockie from './Blockie.js'
 import { Button } from './Button.js'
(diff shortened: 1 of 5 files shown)


app/ts/store/account.ts
```diff
@@ -1,87 +1,86 @@
...
 					assertsExternalProvider(window.ethereum)
 					const provider = new ethers.providers.Web3Provider(window.ethereum)
 					const signer = provider.getSigner()
-					const address = await signer.getAddress()
-					accountStore.value = { isConnected: true, address }
-				} catch {
-					accountStore.value = accountDefaults
+					return await signer.getAddress()
+				} catch (error) {
+					throw new ConnectAttemptError()
```
KillariDev, line 38: we could try to distinguish different errors here and if they don't match the ones we expect, we throw the error at it is

app/ts/store/account.ts
```diff
@@ -1,87 +1,86 @@
...
 
 	const handleAccountChange = (newAccount: string[]) => {
-		if (ethers.utils.isAddress(newAccount[0])) {
-			accountStore.value = { isConnected: true, address: newAccount[0] }
-			return
-		}
-
-		reset()
-		accountStore.value = accountDefaults
+		accountStore.value = ethers.utils.isAddress(newAccount[0]) ? { state: 'connected', address: newAccount[0] } : accountStoreDefaults
```
KillariDev, line 53: the new account list can be empty too

KillariDev (approved):

app/ts/components/ConnectToWallet.tsx
```diff
@@ -10,46 +8,38 @@ import * as Icon from './Icon/index.js'
...
+			const connectToWallet = accountStore.value.connect
 			return (
 				<Wrapper>
 					<div class='grid grid-cols-[minmax(auto,max-content)_minmax(auto,max-content)_minmax(auto,max-content)] gap-3 items-center justify-center'>
 						<div class='leading-tight text-right'>Get started quickly by connecting your wallet</div>
 						<div class='transition animate-bounce-x'>
 							<Icon.ArrowRight />
 						</div>
-						<Button class='whitespace-nowrap' onClick={() => connectMutation.dispatch()}>
+						<Button class='whitespace-nowrap' onClick={() => connectToWallet()}>
```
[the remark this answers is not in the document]
jubalm: there's an optional parameter on `connectToWallet` that TS will complain if i place it directly on `onClick`.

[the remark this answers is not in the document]
MicahZoltu: If you make `connectToWallet` an overloaded function does that resolve the problem?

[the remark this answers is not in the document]
jubalm: `connectToWallet` is just a reference to a function with signature `(attemptOnly?: boolean) => void` so JSX `onClick` complains incompatible type when I used it as a parameter. I did try changing the reference function to `(attemptOnly?:boolean) => void | (() => void)` (from what i understand about overriding in TS) but didn't like it either.

Using `bind` no longer gets complaints from TS; as in `connect.bind(null, false)` and i can even use it directly without creating the pointer.

app/ts/components/ConnectToWallet.tsx
```diff
@@ -24,15 +24,14 @@ export const ConnectToWallet = () => {
...
-			const connectToWallet = accountStore.value.connect
 			return (
 				<Wrapper>
 					<div class='grid grid-cols-[minmax(auto,max-content)_minmax(auto,max-content)_minmax(auto,max-content)] gap-3 items-center justify-center'>
 						<div class='leading-tight text-right'>Get started quickly by connecting your wallet</div>
 						<div class='transition animate-bounce-x'>
 							<Icon.ArrowRight />
 						</div>
-						<Button class='whitespace-nowrap' onClick={() => connectToWallet()}>
+						<Button class='whitespace-nowrap' onClick={accountStore.value.connect.bind(null, false)}>
```
MicahZoltu, line 34: For the record, I liked `() => accountStore.value.connect()` better than the `bind` solution.  However, I'm fine with all of the options, so pick the one you like best.

I think `undefined` for first parameter of `bind` is more correct than `null` though?
```suggestion
						<Button class='whitespace-nowrap' onClick={accountStore.value.connect.bind(undefined, false)}>
```

app/ts/store/account.ts
jubalm: this will be handled in #70

app/ts/components/ConnectToWallet.tsx
jubalm: i'm a fan of arrow functions in event handlers as well but for some reason the type for `accoutStore.value` seem to have lost context as a return value `() => accountStore.value.connect()` in this switch statement.
Screenshot 2023-02-13 at 2 43 11 AM

app/ts/store/account.ts
jubalm: The try-catch here is an attempt to verify if a wallet account is accessible and is expected to withdraw silently if the attempt failed to assume no account is connected yet. nevertheless, `ConnectAttemptError` will now receive the error message

app/ts/components/ConnectToWallet.tsx
MicahZoltu: Ah, yeah.  that makes sense.  This is why you needed the local, that makes sense.

PR #70, opened 2023-02-09
Eth Balance Context and Component

jubalm:
Dependent of #69 

Rework of #59 using new async states

Notable changes:
- state design similar to account
- add account change balance update
- directory and file structure (discussion)
  - the code in a `context` folder within a single file as they all work together and requires each other to function.

### DIFF
--- a/app/ts/components/EthBalance.tsx
+++ b/app/ts/components/EthBalance.tsx
@@ -1,32 +1,36 @@
 import { ethers } from 'ethers'
 import { assertUnreachable } from '../library/utilities.js'
 import { useBalanceStore } from '../context/BalanceContext.js'
-import { useAccountStore } from './AccountContext.js'
 
 export const EthBalance = () => {
-	const accountStore = useAccountStore()
 	const balanceStore = useBalanceStore()
 
-	if (accountStore.value.state === 'disconnected') {
-		const connectToWalet = accountStore.value.connect
-		return (
-			<div class='flex flex-col items-center justify-center md:items-start md:gap-1 mb-2'>
-				<div class='uppercase text-white/50 text-xs md:text-sm flex'>balance (ETH)</div>
-				<div class='flex items-center gap-2'>
-					<div>
-						<button class='font-bold text-white/80 hover:text-white' onClick={() => connectToWalet()}>
-							Connect wallet
-						</button>{' '}
-						<span class='text-white/50'>to view balance</span>
+	switch (balanceStore.value.state) {
+		case 'unavailable':
+			return (
+				<div class='flex flex-col items-center justify-center md:items-start md:gap-1 mb-2'>
+					<div class='uppercase text-white/50 text-xs md:text-sm flex'>balance (ETH)</div>
+					<div class='flex items-center gap-2'>
+						<div>Balance is currently unavailable</div>
 					</div>
 				</div>
-			</div>
-		)
-	}
-
-	switch (balanceStore.value.state) {
-		case 'initial':
-			balanceStore.value.checkBalance()
+			)
+		case 'disconnected':
+			const connectWallet = balanceStore.value.connect
+			return (
+				<div class='flex flex-col items-center justify-center md:items-start md:gap-1 mb-2'>
+					<div class='uppercase text-white/50 text-xs md:text-sm flex'>balance (ETH)</div>
+					<div class='flex items-center gap-2'>
+						<div>
+							<button class='font-bold text-white/80 hover:text-white' onClick={() => connectWallet()}>
+								Connect wallet
+							</button>{' '}
+							<span class='text-white/50'>to view balance</span>
+						</div>
+					</div>
+				</div>
+			)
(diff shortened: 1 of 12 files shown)


jubalm:
Opened for review so can start aligning if there's anything I might have overlooked

app/ts/components/EthBalance.tsx
```diff
@@ -0,0 +1,72 @@
...
+	const balanceStore = useBalanceStore()
+
+	if (accountStore.value.state === 'disconnected') {
+		const connectToWalet = accountStore.value.connect
+		return (
+			<div class='flex flex-col items-center justify-center md:items-start md:gap-1 mb-2'>
+				<div class='uppercase text-white/50 text-xs md:text-sm flex'>balance (ETH)</div>
+				<div class='flex items-center gap-2'>
+					<div>
+						<button class='font-bold text-white/80 hover:text-white' onClick={() => connectToWalet()}>
```
KillariDev, line 17:
```suggestion
						<button class='font-bold text-white/80 hover:text-white' onClick={connectToWalet}>
```

app/ts/components/EthBalance.tsx
```diff
@@ -0,0 +1,72 @@
...
+		case 'failed':
+			const checkBalance = balanceStore.value.reset
+			return (
+				<div class='flex flex-col items-center justify-center md:items-start md:gap-1 mb-2'>
+					<div class='uppercase text-white/50 text-xs md:text-sm flex'>balance (ETH)</div>
+					<div class='flex items-center gap-2'>
+						<div>
+							<span class='text-white/50'>Unable to view balance</span>{' '}
+							<button class='font-bold text-white/80 hover:text-white' onClick={() => checkBalance()}>
+								Retry?
```
KillariDev, line 62: couldn't this be automatic? We should be querying the balance regularly anyay

app/ts/context/BalanceContext.tsx
```diff
@@ -0,0 +1,94 @@
...
+export type BalanceStore = Signal<Balance>
+export function createBalanceStore() {
+	const { value: query, waitFor, reset } = useAsyncState<BigNumber>()
+
+	const checkBalance = () => {
+		waitFor(async () => {
+			assertsExternalProvider(window.ethereum)
+			const provider = new ethers.providers.Web3Provider(window.ethereum)
+			const signer = provider.getSigner()
+			const address = await signer.getAddress()
```
KillariDev, line 37: I think this address should be stored elsewhere, as when you are showing the address, this balance can get out of sync when user switches address, as this retrieves the balance of signers address, and not the one showed in the application

app/ts/context/BalanceContext.tsx
```diff
@@ -0,0 +1,94 @@
...
+			return await provider.getBalance(address)
+		})
+	}
+
+	const balanceStoreDefaults = { state: 'initial' as const, checkBalance }
+	const balanceStore = useSignal<Balance>(balanceStoreDefaults)
+
+	const handleAccountChange = (newAccount: string[]) => {
+		if (ethers.utils.isAddress(newAccount[0])) {
```
KillariDev, line 46: this can result in reading out of bounds if newAccount is empty

app/ts/context/BalanceContext.tsx
```diff
@@ -0,0 +1,94 @@
...
+			checkBalance()
+		}
+	}
+
+	const listenForAccountChanges = () => {
+		assertsExternalProvider(window.ethereum)
+		const provider = new ethers.providers.Web3Provider(window.ethereum)
+		if (!isEthereumObservable(provider.provider)) return
+		provider.provider.on('accountsChanged', handleAccountChange)
```
KillariDev, line 55: remember to remove this listener when its no longer needed

KillariDev (approved):

app/ts/components/EthBalance.tsx
jubalm: New commit b848e5f05dfd8473c56c300e6e7b63469ae571a5 now has automatic balance on block change. This code should at least allow the user to attempt to retry if for some weird reason the wallet decides to throw.

app/ts/context/BalanceContext.tsx
jubalm: replaced this with a `connect` function similar to how one would connect the first time. this approach had issues where a state change is expected by other conditions. Ex. the balance functions need to know the account changed without filtering the signal values as the state will have stayed 'connected' and only the address changed.

app/ts/context/BalanceContext.tsx
jubalm: The address is now shared and should be synced with address set during account connection and account change events.

app/ts/context/BalanceContext.tsx
jubalm: The updated approach no longer relies on these events and instead directly referenced from the address established during connection, ensuring they are synced throughout changes.

PR #72, opened 2023-02-14
Create an application-wide Ethers.js provider

jubalm:
Calling Ethers.js `new ethers.providers.Web3Provider` everytime results in the `chainId` to be out of sync when updating from the connected wallet. Adding this layer of context will ensure that all ether function calls within the context is using the same Provider throughout the app.


app/ts/library/utilities.ts
```diff
@@ -26,25 +27,29 @@ export function removeNonStringsAndTrim(...strings: (string | boolean | undefine
...
 	return ethereum instanceof Object && 'on' in ethereum && typeof ethereum.on === 'function'
 }
 
+export function assertsEthereumObservable(ethereum: unknown): asserts ethereum is ObservableEthereum {
+	if (!isEthereumObservable(ethereum)) throw new Error('Ethereum object is not observable')
+}
+
 export function assertUnreachable(value: never): never {
-	throw new Error(`Never gonna give you up (${value})`)
+	throw new Error(`Unexpected code execution (${value})`)
```
KillariDev, line 52: meme got old? :D

KillariDev (approved):
image

This text does not change even if network changes.

Also when you change network, the balance blinks (because it goes back to "loading" state and then quickly back to value). I think it would be better if it just updates to the new value without going via "loading" state, then it would not blink

app/ts/library/utilities.ts
jubalm: I didn't even realize this got into the codebase until I bumped into one 😆

MicahZoltu: So you are saying past Jubal rickrolled present Jubal?  Well done.

jubalm:
> image
> 
> This text does not change even if network changes.
> 
> Also when you change network, the balance blinks (because it goes back to "loading" state and then quickly back to value). I think it would be better if it just updates to the new value without going via "loading" state, then it would not blink

Addressing in a separate PR

PR #79, opened 2023-03-01
Add balances to asset listing

jubalm:


app/ts/library/utilities.ts
```diff
@@ -24,6 +24,14 @@ export function removeNonStringsAndTrim(...strings: (string | boolean | undefine
 	)
 }
 
+/**
+ * Generate a number within a number range
+ */
+
+export function getRandomNumberBetween(from: number, to: number): number {
```
KillariDev, line 31: why do we need a random number generator?

KillariDev (approved):
Maybe bit too excessive to complain about the balance or every token? Could just leave it empty
image

app/ts/components/AsyncText.tsx
```diff
@@ -0,0 +1,17 @@
+import { ComponentChildren } from 'preact'
+import { getRandomNumberBetween, removeNonStringsAndTrim } from '../library/utilities.js'
+
+export type AsyncTextProps = {
+	placeholderLength?: number
+	children?: ComponentChildren
+	class?: string
+}
+
+export const AsyncText = ({ placeholderLength = getRandomNumberBetween(10, 20), children, class: classNames }: AsyncTextProps) => {
```
0xjimmy, line 10: If you make the placeholderLength required then you don't need to generate random amount.
Does the length generated also not affect how the design looks? I would assume setting a fixed amount would be better to control the design.

app/ts/library/utilities.ts
jubalm: > Maybe bit too excessive to complain about the balance or every token? Could just leave it empty
> image

Moving the changeset to #84

jubalm: > why do we need a random number generator?

Just used in text with unknown length while loading to give it a "presumed" look

app/ts/components/AsyncText.tsx
jubalm: Would be going with this on #84

PR #89, opened 2023-03-14
Add transfer token values to transaction page

### LINKED ISSUE #87: Add token information on transaction details page

jubalm:
- changed erc20 abi passed to ethers to string functions (maybe later pull from a dictionary instead)
- reworked transaction context
- added checks for transactions validating token transfers

Resolves #87


app/ts/library/ethereum.ts
```diff
@@ -0,0 +1,22 @@
...
+	return txResponse.data.startsWith('0xa9059cbb')
+}
+
+export function getTransferTokenValue(transactionReceipt: TransactionReceipt): BigNumber {
+	const IERC20 = new ethers.utils.Interface(ERC20ABI)
+
+	let tokenValue = undefined
+
+	for (const receiptLog of transactionReceipt.logs) {
+		const parsedLog = IERC20.parseLog(receiptLog)
```
MicahZoltu, line 15: What happens if `parseLog` fails?  Does it throw or return something else?  It is very possible that there are multiple events in a transfer, some of which may not be parsable with `ERC20ABI`.

app/ts/library/ethereum.ts
```diff
@@ -0,0 +1,22 @@
...
+		if (parsedLog.name.toLowerCase() === 'transfer') {
+			tokenValue = parsedLog.args.value
+		}
```
MicahZoltu, line 18: This appears to use the value of the last token transfer in a transaction.  Why not the first?

Even better would be for this function to take the expected sender, recipient, and token and make sure the transfer log matches everything and only extract the value when the rest match.

Alternatively, this function could return *all* of the parsed transfer logs and leave it up to the caller to decide which one to use in the case of multiple.

app/ts/library/ethereum.ts
```diff
@@ -0,0 +1,22 @@
+import { BigNumber, ethers } from 'ethers'
+import { TransactionReceipt, TransactionResponse, TransferTransactionResponse } from '../types.js'
+import { ERC20ABI } from './ERC20ABI.js'
+
+export function isTransferTransaction(txResponse: TransactionResponse): txResponse is TransferTransactionResponse {
+	return txResponse.data.startsWith('0xa9059cbb')
```
MicahZoltu, line 6: Just in case, it would be best to do a case insensitive match here.
```suggestion
	return txResponse.data.toLower().startsWith('0xa9059cbb')
```

MicahZoltu (approved):

app/ts/library/ethereum.ts
jubalm: Adjusted in
https://github.com/DarkFlorist/lunaria/pull/89/commits/540795b94176d74702c283d631cf134bc1aa761a. Now it should return `undefined` to hand over handling to the next call

app/ts/library/ethereum.ts
jubalm: i used a different approach after https://github.com/DarkFlorist/lunaria/pull/89/commits/540795b94176d74702c283d631cf134bc1aa761a

> This appears to use the value of the last token transfer in a transaction. Why not the first?

I'm not aware of this scenario so I replaced the loop with `find` so that only the first matched log will be used

Added type guards as well to limit response to only the expected transfer log result

app/ts/library/ethereum.ts
```diff
@@ -0,0 +1,32 @@
...
+	const erc20Interface = new ethers.utils.Interface(ERC20ABI)
+	const transferLog = transactionReceipt.logs.find(isTransferLog)
+	if (transferLog === undefined) return undefined
+	const logArgs = erc20Interface.parseLog(transferLog).args
+	return isTransferResult(logArgs) ? logArgs.value : undefined
+}
+
+export function isTransferLog(log: ethers.providers.Log) {
+	const [topic] = log.topics
+	const transferTopic = ethers.utils.id('Transfer(address,address,uint256)')
```
MicahZoltu, line 20: This is a *little* bit expensive to calculate (involves a hash).  Consider moving this up to a file-local variable (up 3 lines) so it is calculated once rather than every time this function is called.

app/ts/library/ethereum.ts
```diff
@@ -0,0 +1,32 @@
...
+import { TransactionReceipt, TransactionResponse, TransferTransactionResponse } from '../types.js'
+import { ERC20ABI } from './ERC20ABI.js'
+
+export function isTransferTransaction(txResponse: TransactionResponse): txResponse is TransferTransactionResponse {
+	return txResponse.data.toLowerCase().startsWith('0xa9059cbb')
+}
+
+export function getTransferTokenValue(transactionReceipt: TransactionReceipt) {
+	const erc20Interface = new ethers.utils.Interface(ERC20ABI)
+	const transferLog = transactionReceipt.logs.find(isTransferLog)
```
MicahZoltu, line 12: I'm still a bit uncomfortable with this.  A single transaction can have multiple transfer events.  For example, some ERC20 tokens will transfer a fee to the token creator every time a token is transferred, as well as transferring to the designated recipient.  In this code we are blindly assuming that the first transfer event (log) is the one we care about, but there is no guarantee in the standard that that is the case.

Do we know who the expected sender/recipient is when we are calling this?  If so, we should filter on those.  If we don't, we may want to consider just presenting all of the events to the user rather than just the first one.

Note: If it is a lot of work to do these, we can file issues for them and address them in a separate PR.

MicahZoltu (approved):

app/ts/library/ethereum.ts
```diff
@@ -0,0 +1,32 @@
+import { BigNumber, ethers } from 'ethers'
+import { Result } from 'ethers/lib/utils.js'
+import { TransactionReceipt, TransactionResponse, TransferTransactionResponse } from '../types.js'
+import { ERC20ABI } from './ERC20ABI.js'
+
+export function isTransferTransaction(txResponse: TransactionResponse): txResponse is TransferTransactionResponse {
+	return txResponse.data.toLowerCase().startsWith('0xa9059cbb')
```
0xjimmy, line 7: I don't see this used anywhere but it will only detect `transfer` calls from an EOA. So that doesn't include contract interactions that use `transferFrom` internally.

app/ts/library/ethereum.ts
0xjimmy: Yes I agree, should definitely find all matching transfer.

0xjimmy (approved):

PR #92, opened 2023-03-28
Assets page updated build

### LINKED ISSUE #71: Add navigation to send page and transaction pages

### LINKED ISSUE #90: Refine log parsing in gathering transaction details
I'm still a bit uncomfortable with this.  A single transaction can have multiple transfer events.  For example, some ERC20 tokens will transfer a fee to the token creator every time a token is transferred, as well as transferring to the designated recipient.  In this code we are blindly assuming that the first transfer event (log) is the one we care about, but there is no guarantee in the standard that that is the case.

Do we know who the expected sender/recipient is when we are calling this?  If so, we should filter on those.  If we don't, we may want to consider just presenting all of the events to the user rather than just the first one.

Note: If it is a lot of work to do these, we can file issues for them and address them in a separate PR.

_Originally posted by @MicahZoltu in https://github.com/DarkFlorist/lunaria/pull/89#discussion_r1140943571_

jubalm:
Currently this PR is dependent on
- [ ] #95 will need to be rebased to this after merge 

This PR will have resolved a couple more issues

- [ ] resolve #90
- [ ] resolve #71

### DIFF
--- a/app/css/index.css
+++ b/app/css/index.css
@@ -856,6 +856,11 @@ video {
 	border-color: rgb(255 255 255 / 0.1);
 }
 
+.border-gray-200 {
+	--tw-border-opacity: 1;
+	border-color: rgb(229 231 235 / var(--tw-border-opacity));
+}
+
 .border-white\/20 {
 	border-color: rgb(255 255 255 / 0.2);
 }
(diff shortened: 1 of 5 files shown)


jubalm:
Close in favor of #111

PR #94, opened 2023-03-30
Add notify component for error handling

jubalm:
Component will be used to catch errors and display to user

### DIFF
--- a/app/ts/components/Notice.tsx
+++ b/app/ts/components/Notice.tsx
@@ -6,10 +6,8 @@ export const Notices = () => {
 	if (notices.value.length < 1) return <></>
 
 	return (
-		<>
-			<Modal title={notices.value[0]?.title} onClose={() => (notices.value = notices.value.filter(n => n.id !== notices.value[0]?.id))} hasCloseButton>
-				{notices.value[0]?.message}
-			</Modal>
-		</>
+		<Modal title={notices.value[0]?.title} onClose={() => (notices.value = notices.value.filter(n => n.id !== notices.value[0]?.id))} hasCloseButton>
+			{notices.value[0]?.message}
+		</Modal>
 	)
 }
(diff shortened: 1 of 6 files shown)


app/ts/components/Notice.tsx
```diff
@@ -0,0 +1,15 @@
...
+export const Notices = () => {
+	const { notices } = useNotice()
+	if (notices.value.length < 1) return <></>
+
+	return (
+		<>
+			<Modal title={notices.value[0]?.title} onClose={() => (notices.value = notices.value.filter(n => n.id !== notices.value[0]?.id))} hasCloseButton>
+				{notices.value[0]?.message}
+			</Modal>
+		</>
```
MicahZoltu, line 13:
```suggestion
		<Modal title={notices.value[0]?.title} onClose={() => (notices.value = notices.value.filter(n => n.id !== notices.value[0]?.id))} hasCloseButton>
			{notices.value[0]?.message}
		</Modal>
```

app/ts/store/notice.ts
```diff
@@ -0,0 +1,18 @@
...
+export type Notice = {
+	id: number
+	title: string
+	message: string
+}
+
+const notices = signal<Notice[]>([])
+
+export function useNotice() {
+	const notify = (notice: Optional<Notice, 'id'>) => {
```
MicahZoltu, line 13: It feels like `notify` should not allow caller provided `id` because it is very possible/likely that two callers will use the same ID.  by having this function manage IDs we don't have to worry about collisions.

On a related note, instead of `notices.value.length + 1` we should just have a file-local `let id = 0` and use `++id` for the ID of each notice.  This way they never get reused (which could cause some problems).

KillariDev (approved):

app/ts/store/notice.ts
jubalm: removed option for notice id and signal computed the id instead

PR #95, opened 2023-03-30
Updated store providers

jubalm:
Updates here are part of a larger PR and contains only functionality. The UI is on another PR. There is a demo page in `#test` route to test the new store.

the Demo page will be removed before merging this PR.


MicahZoltu (changes requested):
I'll give this another look after the base branch issue has been resolved, just request another review.

app/ts/components/App.tsx
```diff
@@ -5,28 +5,23 @@ import { EthereumProvider } from '../context/EthereumProvider.js'
 import { AccountProvider } from '../context/Account.js'
 import { TransferPage } from './TransferPage/index.js'
 import { TransactionDetails } from './TransactionDetailsPage.js'
-import { Notices } from './Notice.js'
-import { useNotice } from '../store/notice.js'
+import { Lab } from './Lab.js'
 
 export function App() {
-	const { notify } = useNotice()
```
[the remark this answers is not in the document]
jubalm: this PR depends on that branch #94 so i intentionally pointed it there until that one gets merged so i don't accidentally push to master.

MicahZoltu (changes requested):
Page doesn't load at all if you have no wallet installed:
> Uncaught Error: Wallet not installed or incompatible.

app/ts/components/Demo.tsx
```diff
@@ -0,0 +1,142 @@
...
+		if (name === 'chainId' || name === 'decimals') {
+			inputValue = parseInt(inputValue)
+		}
+
+		tokenMeta.value = { ...tokenMeta.value, [name]: inputValue }
+	}
+
+	const handleSubmit = (e: JSX.TargetedEvent<HTMLFormElement>) => {
+		e.preventDefault()
+		addToken(tokenMeta.value)
```
KillariDev, line 128: do we check that the tokens are valid? eg, address is a proper address? We could also check that the token exists, decimals/symbols could also be fetched from chain

app/ts/components/Demo.tsx
```diff
@@ -0,0 +1,142 @@
...
+		return network.value.state !== 'resolved' ? 1 : network.value.value.chainId
+	})
+
+	const tokenMeta = useSignal<TokenMeta>({
+		chainId: chainId.value,
+		name: '',
+		address: '',
+		symbol: '',
+		decimals: 18,
+	})
```
KillariDev, line 113: should we have icon too?

app/ts/store/tokens.ts
```diff
@@ -0,0 +1,112 @@
...
+
+const retrieveTokensFromCache = (cacheKey: string) => {
+	const tokensCache = localStorage.getItem(cacheKey)
+
+	if (tokensCache === null) {
+		return DEFAULT_TOKENS
+	}
+
+	try {
+		return JSON.parse(tokensCache) as TokenMeta[]
```
KillariDev, line 14: I think this should be properly parsed and not casted

app/ts/store/tokens.ts
```diff
@@ -0,0 +1,112 @@
...
+	} catch {
+		throw new Error('Tokens cache may be corrupted.')
```
KillariDev, line 16:
```suggestion
		throw new Error('Token cache is corrupted.')
```

app/ts/store/tokens.ts
```diff
@@ -0,0 +1,112 @@
...
+		tokens,
+		addToken,
+	}
+}
+
+// Move to constants later
+export type TokenMeta = {
+	chainId: number
+	name: string
+	address: string
```
KillariDev, line 57: I think it would be better if these are bigints when parsed from storage and then stored as strings to storage

app/ts/store/tokens.ts
```diff
@@ -0,0 +1,112 @@
...
+	symbol: string
+	decimals: number
+}
+
+export const DEFAULT_TOKENS: TokenMeta[] = [
+	{
+		chainId: 1,
+		address: '0x4c9bbfc1fbd93dfb509e718400978fbeedf590e9',
+		name: 'Rai',
```
KillariDev, line 66:
```suggestion
		name: 'Rai Reflex Index',
```

app/ts/components/Demo.tsx
jubalm: there is a separate component later to handle which icon to load. they will basically be basically be images files named from from addresses.

app/ts/components/Demo.tsx
jubalm: i updated the demo for the add token, including pulling information from the Contract instance and calling their actual meta.

app/ts/store/tokens.ts
jubalm: i now updated the function to validate parsed json from cache including assertions to narrow down the resulting array

jubalm:
> Page doesn't load at all if you have no wallet installed:
> 
> > Uncaught Error: Wallet not installed or incompatible.

There is a missing condition in checking account state listener which should be fixed in https://github.com/DarkFlorist/lunaria/pull/95/commits/5e15f46f74d76b384fb89d52e9f38613e02a31ff

app/ts/store/tokens.ts
jubalm: this change may break so i created #102 to address later

app/ts/components/Demo.tsx
jubalm: created a new PR for the icon strategy #103

app/ts/components/Demo.tsx
```diff
@@ -0,0 +1,139 @@
...
+const Address = () => {
+	const { address, connect } = useAccount()
+
+	switch (address.value.state) {
+		case 'inactive':
+			return <Button text='Connect' onClick={() => connect()} />
+		case 'pending':
+			return <>connecting...</>
+		case 'rejected':
+			return <>error</>
```
MicahZoltu, line 39: I currently just see `error` after clicking connect.  No details provided in the log or on the page.  Perhaps here we could just do:
```suggestion
		case 'rejected':
			return <>error: ${address.value.error.message}</>
```

MicahZoltu (dismissed):
Getting closer to working (page now loads) but I cannot connect with interceptor attached (no MetaMask).

app/ts/components/Demo.tsx
jubalm: looks like `window.ethereum` does not have a `addListener` function if you are using other than MetaMask to connect. I've changed it to `on` as I found that Brave also supports this.

FWIW: the function seem to vary between vendor and I doesn't seem to be on 1193 specs

app/ts/store/provider.ts
```diff
@@ -0,0 +1,45 @@
...
+	// reinitialize provider
+	assertsWithEthereum(window)
+	provider.value = new ethers.providers.Web3Provider(window.ethereum)
+}
+
+const removeChainChangeListener = effect(() => {
+	if (provider.value === undefined) return
+	assertsWithEthereum(window)
+	window.ethereum.on('chainChanged', handleChainChange)
+})
```
0xjimmy, line 45: When network changes you don't need to reinitialize the provider, for a future optimization you could just make sure to remove event listener on disconnect instead.
Right now you, set the value of the provider store again, which triggers Signal effect in `app/ts/store/network.ts` which then queries the network from the provider.
You could just use the network provided in the network change event, this would remove the need for a signalEffect and rpc calls to fetch network.

0xjimmy (approved):

app/ts/store/provider.ts
jubalm: Will take a look closer as a separate issue #106

PR #103, opened 2023-04-18
Token List and Token Icons Handling

jubalm:
Changes on this PR do not yet have a direct link but can be accessed through `#assets` for preview.

Depends on #95


app/ts/components/AssetsPage/AddTokenDialog.tsx
```diff
@@ -0,0 +1,32 @@
...
+			<div class='relative bg-white w-full max-w-md'>
+				<div class='grid justify-end'>
+					<button class='h-10 w-10' onClick={handleCancel}>
+						&times;
+					</button>
+				</div>
+				<div class='px-6 py-6'>
+					<div class='font-bold mb-2'>Add Token</div>
+					<div class='mb-2'>Enter the token's contract address to retrieve details</div>
+					<QueryToken />
```
MicahZoltu, line 27: I just get an error when I put in a token address (like DAI):
> The active network failed to retrieve token information.
This page doesn't appear to have any connect button.  I tried connecting on the main page (to interceptor) and then navigating to #assets but I still get the error.

MicahZoltu (changes requested):
I would like to see the test page working before approving this, since it doesn't yet show up anywhere in the app.

I do like the idea of creating little hidden test pages like this by the way!

KillariDev (approved):
when I add token, it briefly flash that its added and then the token disappears
image

- currently you add tokens as .svg:s, generally token lists have various token image formats that need to be supported later, and we don't get nice svgs. Or how did you get these svgs?

- search doesn't search ticker
- I don't think its obvious that the first box is search box, I guess it needs icon?

PR #104, opened 2023-04-25
General Page Layout

jubalm:
This will be the layout structure for Assets page with a demo on `#layout` route. Notably need a review on other devices than mac and should not have the scrollbar artifacts previously noticed on assets listing.

[UPDATE]: remove demo pages


0xjimmy:
Use `overflow-auto` instead if `overflow-scroll`, as chromium renders empty scroll bars even if the content is too small to scroll.

app/ts/store/tokens.ts
```diff
@@ -0,0 +1,174 @@
...
+	chainId: number
+	name: string
+	address: string
+	symbol: string
+	decimals: number
+}
+
+export const DEFAULT_TOKENS: TokenMeta[] = [
+	{
+		chainId: 1,
```
KillariDev, line 96: image is missing from here?

KillariDev (approved):
This adds token images again like the previous PR?

app/ts/store/tokens.ts
jubalm: image icon will be loaded as publicPath + lowecased contract address + file extension else will show default ethereum icon

jubalm:
> This adds token images again like the previous PR?

rebased to remove confusion

MicahZoltu:
This PR just removed the demo pages and nothing, why was that?  I found the demo pages to be very valuable.

PR #110, opened 2023-05-06
Create .npmrc

jubalm:
NPM configurations added
- [x] configure npm to ingore scripts on build
- [x] configure npm to install exact versions on install


MicahZoltu (approved):

MicahZoltu:
Good idea, we should get this in all of our projects cc @KillariDev @0xjimmy

PR #113, opened 2023-05-11
Re-implement connect component

jubalm:
Separated the functions for the new UI into smaller PRs. This will have included
- [x] apply connect functionality to new ui
- [x] remove random length to loading text
- [x] move network below wallet address

### DIFF
--- a/app/css/index.css
+++ b/app/css/index.css
@@ -616,6 +616,10 @@ img,
   display: flex;
 }
 
+.inline-flex {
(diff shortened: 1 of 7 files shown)


0xjimmy (approved):
You could make the pending shapes the same size and margins as the content. They don't line up the same
image
image
I would expect the skeleton blockie to be the same height as the address and network skeletons.

PR #123, opened 2023-05-25
Add balances to token card

### LINKED ISSUE #117: Re-add token balances in Token selection
- [x] create PR
- [ ] get reviews for PR #123

jubalm:
Added token balance to token cards

Resolves #117

### DIFF
--- a/app/ts/components/TokenManager/EtherBalance.tsx
+++ b/app/ts/components/TokenManager/EtherBalance.tsx
@@ -1,0 +1,50 @@
+
+import { useSignalEffect } from '@preact/signals'
+import { BigNumber, ethers } from 'ethers'
+import { useAsyncState } from '../../library/preact-utilities.js'
+import { useAccount } from '../../store/account.js'
+import { useNetwork } from '../../store/network.js'
+import { useProviders } from '../../store/provider.js'
+import { AsyncText } from '../AsyncText.js'
+
+export const EtherBalance = () => {
+	const { network } = useNetwork()
+	const { address } = useAccount()
+	const providers = useProviders()
+	const { value: query, waitFor } = useAsyncState<BigNumber>()
+
+	const getBalance = () => {
+		if (address.value.state !== 'resolved') return
+		const accountAddress = address.value.value
+		waitFor(async () => {
+			const provider = providers.getbrowserProvider()
+			return await provider.getBalance(accountAddress)
+		})
+	}
+
+	useSignalEffect(() => {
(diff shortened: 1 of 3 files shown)


0xjimmy (approved):
Doesn't show Ether balance, might want to create an issue for that later.

jubalm:
> Doesn't show Ether balance, might want to create an issue for that later.

True, i was going to add to another PR but code wasn't significant so just added the commit here. [c58dfa9](https://github.com/DarkFlorist/lunaria/pull/123/commits/c58dfa986012a166ea98bb3eca60f5e36468906f)

app/ts/components/TokenManager/EtherBalance.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { useAccount } from '../../store/account.js'
+import { useNetwork } from '../../store/network.js'
+import { useProviders } from '../../store/provider.js'
+import { AsyncText } from '../AsyncText.js'
+
+export const EtherBalance = () => {
+	const { network } = useNetwork()
+	const { address } = useAccount()
+	const providers = useProviders()
+	const { value: query, waitFor } = useAsyncState<BigNumber>()
```
MicahZoltu, line 14: Out of curiosity, why do you name it `query` here?

app/ts/components/TokenManager/EtherBalance.tsx
```diff
@@ -0,0 +1,50 @@
...
+		waitFor(async () => {
+			const provider = providers.getbrowserProvider()
+			return await provider.getBalance(accountAddress)
+		})
+	}
+
+	useSignalEffect(() => {
+		if (network.value.state !== 'resolved' || address.value.state !== 'resolved') return
+		getBalance()
+	})
```
MicahZoltu, line 28: Consider inlining `getBalance` into the `useSignalEffect`.  This would save you from having to check `address.value.state` twice.  Alternatively, if you like having a named function consider moving 100% of the `useSignalEffect` logic into `getBalance`.

MicahZoltu: Same comment for token balance component below.

MicahZoltu (approved):

app/ts/components/TokenManager/EtherBalance.tsx
jubalm: the keyword `value` can sometimes get disorientating? i sometimes couldn't differentiate at a glance if it's a signal or asyncState. maybe worth naming it something else by default?

FWIW: the keyword `query` was from a habit formed from using `react-query` library before 😬

PR #125, opened 2023-05-27
Re-add custom token functionality

jubalm:
Migrated the test page components onto the new UI


app/ts/components/DefaultLayout/AddTokenModal.tsx
```diff
@@ -1,16 +1,33 @@
-import { ComponentChildren } from "preact"
+import { Signal } from '@preact/signals'
+import { QueryToken } from '../QueryToken.js'
 
-export type AddTokenModalProps = {
-	children: ComponentChildren
-	onClose: () => void
+type Props = {
+	show: Signal<'select' | 'add' | undefined>
```
MicahZoltu: I have an `OptionalSignal` that you may find useful: https://github.com/MicahZoltu/ethereum-toolbox-ui/blob/main/app/ts/library/preact-utilities.ts#L52-L99

It provides a signal that lets you extract out a defined version of the underlying signal.  This reduces re-rendering when you have code like:
```ts
if (mySignal.value !== undefined) <MyOtherComponent signalProp={mySignal.value}/>
```
In this example, MyOtherComponent receives a signal rather than receiving a value, which means `MyOtherComponent` can pass the signal down to its children without `MyOtherComponent` re-rendering.

When you don't want this feature, you can use `mySignal.deepValue` which will return T | undefined.

MicahZoltu (approved):

PR #132, opened 2023-06-03
Transaction Page Balance

### LINKED ISSUE #118: Show balance after transfer success

jubalm:
Resolves #118


app/css/index.css
MicahZoltu, line 1: Indentation changed in this file again.  You may want to investigate why this is happening so whatever is causing it can be updated/fixed.

app/ts/components/TransactionPage/TransactionDetails.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+	// automatically get transaction receipt
+	useSignalEffect(() => {
+		if (transactionResponse.value.state !== 'resolved') return
+		getTransactionReceipt(transactionResponse.value.value)
+	})
+
+	// reset async states
+	useEffect(() => {
+		resetResponse()
```
MicahZoltu, line 123: This is unnecessary since you call `gettransactionResponse` immediately below.  `useAsyncState` handles all of the issues around firing off multiple requests in a row before receiving a response (it drops the old ones and uses the new one).

MicahZoltu (approved):

app/css/index.css
jubalm: fixed, it was happening when i use the --watch flag as it skips the formatting step

PR #146, opened 2023-06-15
Blockie update

### LINKED ISSUE #129: Update Blockie similar to TheInterceptor
Blockie at TheInterceptor now requires a bit more libraries to work with.
- [x] Get to understand what the changes offer and if Lunaria needs them
- [x] Add blockies to all addresses

Resolved at #146

jubalm:
Resolves #129

- [x] add blockie to address information on transaction page

Screenshot 2023-06-15 at 1 57 02 PM


KillariDev (approved):
blockie is missing from this address:
image

app/ts/components/SVGBlockie.tsx
```diff
@@ -0,0 +1,97 @@
...
+	const spotcolor = createColor()
+	const imageData = createImageData(options.size || 8)
+
+	return { imageData, color, bgcolor, spotcolor }
+}
+
+// SVGBlockie component is resizable through CSS font size
+export default function SVGBlockie({ address }: { address: string }) {
+	const pixelDensity = 8
+	const { imageData, color, spotcolor, bgcolor } = generateIdenticon({ seed: address, size: pixelDensity })
```
0xjimmy, line 85: Could memorize the result of generateIdenticon

0xjimmy (approved):

app/ts/components/SVGBlockie.tsx
MicahZoltu: I'm not sure it is worth bothering here given how cheap this generation is (basically a loop over an 8x8 with no cryptography involved).  However, I think this code would provide memoization if that is desired.
```suggestion
	const { imageData, color, spotcolor, bgcolor } = useMemo(() => generateIdenticon({ seed: address, size: pixelDensity }), [address])
```

PR #163, opened 2023-07-05
Add token image fallback

### LINKED ISSUE #157: When new token is added, it shows missing image
image

jubalm:
Resolves #157 

Tokens that don't have icons on the website library will show this icon as a fallback

Screenshot 2023-07-06 at 3 02 48 PM


app/ts/components/TokenManager/index.tsx
```diff
@@ -113,7 +113,9 @@ const TokenCard = ({ token, onSelect }: TokenCardProps) => {
 	return (
 		<div class='aspect-square bg-neutral-800 p-4 flex items-end relative group outline-none' tabIndex={1}>
 			<div class='relative'>
-				<img class='w-12 h-12 mb-2' src={`/img/${token.address.toLowerCase()}.svg`} />
+				<object class='w-12 h-12 mb-2' data={`/img/${token.address.toLowerCase()}.svg`} type="image/png">
+					<img class='w-12 h-12' src="/img/ethereum.svg" />
+				</object>
```
MicahZoltu, line 118: How does this result in a fallback image?

app/ts/components/TokenManager/index.tsx
```diff
@@ -113,7 +113,9 @@ const TokenCard = ({ token, onSelect }: TokenCardProps) => {
 	return (
 		<div class='aspect-square bg-neutral-800 p-4 flex items-end relative group outline-none' tabIndex={1}>
 			<div class='relative'>
-				<img class='w-12 h-12 mb-2' src={`/img/${token.address.toLowerCase()}.svg`} />
+				<object class='w-12 h-12 mb-2' data={`/img/${token.address.toLowerCase()}.svg`} type="image/png">
+					<img class='w-12 h-12' src="/img/ethereum.svg" />
```
MicahZoltu, line 117: We shouldn't use the ETH logo as a fallback for missing logo.  Consider a question mark or some other indicator of "unknown" to clearly indicate that we don't have an icon.

MicahZoltu (changes requested):
I would like to see something other than the ETH logo as a fallback before this is merged.

app/ts/components/TokenManager/index.tsx
jubalm: an object element can load an external resource, just like how a video behaves, and will fallback to the underlying `` while it fetches or if it failed to fetch the resource.

app/ts/components/TokenManager/index.tsx
jubalm: Replaced the icon with this:
Screenshot 2023-07-06 at 3 02 48 PM

MicahZoltu: Hmm, screenshot missing?

MicahZoltu (approved):

PR #167, opened 2023-07-06
Fix add token functionality

jubalm:
- [x] converted decimals returned from contract to number

calling `contract.decimals()` now returns `bigint` type (previously number) so formatting with `formatUnits` throws and prevents the token from being saved.

alternatively, looked into converting all instances of decimals to bigint but until the `formatUnits` function natively supports it, would be less code to manage to cast it to Number for now

### DIFF
--- a/app/ts/store/tokens.ts
+++ b/app/ts/store/tokens.ts
@@ -74,7 +74,7 @@ export function useTokenQuery() {
 				const name = await contract.name()
 				const symbol = await contract.symbol()
 				const decimals = await contract.decimals()
-				return { chainId, name, symbol, decimals, address: tokenAddress.value } as const
+				return { chainId, name, symbol, decimals: Number(decimals), address: tokenAddress.value } as const
 			} catch (unknownError) {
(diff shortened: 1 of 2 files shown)


app/ts/store/tokens.ts
```diff
@@ -74,7 +74,7 @@ export function useTokenQuery() {
 				const name = await contract.name()
 				const symbol = await contract.symbol()
 				const decimals = await contract.decimals()
-				return { chainId, name, symbol, decimals, address: tokenAddress.value } as const
+				return { chainId, name, symbol, decimals: Number(decimals), address: tokenAddress.value } as const
```
MicahZoltu, line 77: `bigint` is the correct type for `decimals` as they must be integer values, and the contract returns an integer.  We should instead fix the problem of handling `bigint`s elsewhere in the stack rather than converting to `number` here.

MicahZoltu (changes requested):

MicahZoltu (approved):

PR #168, opened 2023-07-06
Remove @node/types and add workaround

jubalm:
A workaround from this template was applied to prevent issues from ethers.js' @node/types dependency when used on a browser environment
https://github.com/Zoltu/preact-es2015-template/blob/master/tsconfig.json#L28-L30

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -59,7 +59,7 @@
 		},
 		"node_modules/@preact/signals-core": {
 			"version": "1.2.1",
-			"resolved": "https://registry.npmjs.org/@preact/signals-core/-/signals-core-1.3.1.tgz",
+			"resolved": "https://registry.npmjs.org/@preact/signals-core/-/signals-core-1.2.1.tgz",
 			"integrity": "sha512-aqzRMNFU1hoQyP4Kb1ldJrUTCnA9vqPDa7qHEQzHJ3upnBOcC2pjmvjAuTqGuY4AVTtUkCQV0FvOCuIQQ2hSdA==",
 			"funding": {
 				"type": "opencollective",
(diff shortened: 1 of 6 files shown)


package-lock.json
```diff
@@ -58,9 +58,9 @@
 			}
 		},
 		"node_modules/@preact/signals-core": {
-			"version": "1.2.1",
-			"resolved": "https://registry.npmjs.org/@preact/signals-core/-/signals-core-1.2.1.tgz",
-			"integrity": "sha512-aqzRMNFU1hoQyP4Kb1ldJrUTCnA9vqPDa7qHEQzHJ3upnBOcC2pjmvjAuTqGuY4AVTtUkCQV0FvOCuIQQ2hSdA==",
+			"version": "1.3.1",
+			"resolved": "https://registry.npmjs.org/@preact/signals-core/-/signals-core-1.3.1.tgz",
+			"integrity": "sha512-DL+3kDssZ3UOMz9HufwSYE/gK0+TnT1jzegfF5rstgyPrnyfjz4BHAoxmzQA6Mkp4UlKe8qjsgl3v5a/obzNig==",
```
MicahZoltu, line 63: This shouldn't have changed without a change to `package.json`.
```suggestion
			"version": "1.2.1",
			"resolved": "https://registry.npmjs.org/@preact/signals-core/-/signals-core-1.3.1.tgz",
			"integrity": "sha512-aqzRMNFU1hoQyP4Kb1ldJrUTCnA9vqPDa7qHEQzHJ3upnBOcC2pjmvjAuTqGuY4AVTtUkCQV0FvOCuIQQ2hSdA==",
```

MicahZoltu (changes requested):
Please rollback signls-core in this PR, since it shouldn't be changing without an update to `package.json`.  Otherwise looks good.

MicahZoltu (approved):

PR #171, opened 2023-07-09
Updated log parsing function

### LINKED ISSUE #124: Fix transaction failed handler

### LINKED ISSUE #162: Fix issue with ethers when resolved
https://github.com/ethers-io/ethers.js/issues/4029

### LINKED ISSUE #170: Unable to save token transfer
Token transfer detail no longer showing the save to favorite form.
- initial diagnosis point to v6 changes in log parsing

jubalm:
Resolves #170 - fixed log parsing causing token save to not show up
Resolves #162 - added forward compatibility for ethers type issue https://github.com/ethers-io/ethers.js/issues/4029
Resolves #124 - ethers v6 update properly returns `null` if it fails to get transaction and is now properly handled

### DIFF
--- a/app/ts/library/ethereum.ts
+++ b/app/ts/library/ethereum.ts
@@ -1,4 +1,4 @@
-import { TransactionResponse, Interface, id, Result, TransactionReceipt, Log, Eip1193Provider, formatEther } from 'ethers'
+import { TransactionResponse, Interface, id, TransactionReceipt, Log, Eip1193Provider, formatEther } from 'ethers'
 import { ERC20ABI } from './ERC20ABI.js'
 import { WalletError } from './exceptions.js'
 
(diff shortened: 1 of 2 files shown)


app/ts/library/ethereum.ts
```diff
@@ -41,28 +41,18 @@ export function isTransferTransaction(txResponse: TransactionResponse): txRespon
...
-}
-
-export interface TransferResult extends Result {
-	from: string
-	to: string
-	value: bigint
+export function extractArgValue<T>({ topics, data }: Log, argKey: string): T | null {
+	// ensure topics conforms to parseLog param type
+	const mutableTopics = topics.map(topic => topic)
+	const parsedLog = erc20Interface.parseLog({ topics: mutableTopics, data })
```
MicahZoltu, line 49: This can be simplified (and I suspect is faster too, but that is insignificant).  You may still want a variable for `mutableTopics` to make it clear to the reader why you are making a copy of the array.
```suggestion
	const parsedLog = erc20Interface.parseLog({ topics: [...topics], data })
```
```suggestion
	const mutableTopics = [...topics]
	const parsedLog = erc20Interface.parseLog({ topics: mutableTopics, data })
```

KillariDev: `erc20Interface.parseLog` will throw if it fails to parse logs. Eg the logs do not conform to erc20, I have this kind of ugly helper function to map the failures to null:
```ts
const parseLogIfPossible = (ethersInterface: ethers.Interface, log: { topics: string[], data: string}) => {
	try {
		return ethersInterface.parseLog(log)
	} catch (error) {
		return null
	}
}
```

KillariDev (approved):

app/ts/library/ethereum.ts
jubalm: > This can be simplified (and I suspect is faster too, but that is insignificant). You may still want a variable for `mutableTopics` to make it clear to the reader why you are making a copy of the array.

i added a helper as @KillariDev mentioned and used the spread operator for topics to conform to parseLog parameters

PR #174, opened 2023-07-13
Better error handling

jubalm:
Created error handling mechanism to extend on. This will help organize known errors that can occur within the app and return a friendlier interface for the user

Development only:
- `#errors` url route exist for simulating the popup only

Update:
- [ ] change modal to a toast-like


jubalm:
Update: the UI component should not block the user from continuing an action (which this currently does) so will need to change the appearance to a toast that will dismiss itself after some time

KillariDev (approved):

PR #187, opened 2023-08-08
Use TS validation library

jubalm:


jubalm:
Close in favor of Funtypes

PR #188, opened 2023-08-08
Dynamic default token icon

jubalm:
used the first two letters of the token name as default icon
Screenshot 2023-08-08 at 1 47 21 PM
etters of the token name


KillariDev (approved):
looks cool!

PR #194, opened 2023-08-17
Update CSS Generation

jubalm:
- replaced npm poststyles script with a postcss plugin
  - supports `--watch` flag to dynamically update css on file changes during development
  ```
  // the double dashes is intentional to pass flags to nested npm script when ran from the root project
  npm run styles -- -- --watch
  ```
  - faster css generation times
- removed custom base styles as it should be now same as Tailwind defaults
  - some vendor prefixes removed as they are now widely supported


twcss/prettifycss/index.js
```diff
@@ -0,0 +1,24 @@
+const parse = require('postcss').parse
```
KillariDev, line 1: why do we have javascript files and not typescript?

jubalm: this plugin is only used for the tailwind css generator to uniformly format the css the same way our typescript files. I don't think postcss accepts typescript modules directly at the moment either and this should not be deployed on the public so just skipped the compile efforts

PR #196, opened 2023-08-22
Feature/typescript funtypes

### LINKED ISSUE #121: Consider introducing funtypes in the codebase
use funtypes here instead of casting

_Originally posted by @KillariDev in https://github.com/DarkFlorist/lunaria/pull/111#discussion_r1203487944_

### LINKED ISSUE #195: Improve UX on handling adding of custom tokens if wallet is not connected
Right now it would not make a lot of sense to add custom token if a user does not have an account connected as storage key is based on account address.

- [ ] prompt user to connect a wallet before showing add token facility

jubalm:
Changeset:

- [x] introduced funtypes to parse localstorage retrieved data - Resolves #121
- [x] add persistent-signal library to sync signal to localstorage
- [x] use new hook managed with funtypes for token list management 
- [x] replaced `add token` button to `connect wallet` when no wallet is connected - Resolves #195 
- [x] change `npm install` scripts to `npm ci --ignore-scripts`
- [x] remove unused types


app/ts/schema.ts
```diff
@@ -0,0 +1,41 @@
...
+		const serializedValue = TokenListSchema.safeSerialize(value)
+		if (!serializedValue.success) return { success: false, message: serializedValue.message }
+		const jsonString = JSON.stringify(serializedValue.value)
+		return { success: true, value: jsonString }
+	},
+}
+
+export const ManagedTokensCacheSchema = funtypes.String.withParser(ManagedTokensCacheParserConfig)
+
+export type ManagedTokensCache = funtypes.Static<typeof ManagedTokensCacheSchema>
```
0xjimmy, line 41:
```suggestion

```
Shouldn't need these, TokenListSchema.serialize() should work fine as is

package.json
```diff
@@ -9,13 +9,13 @@
...
 		"@preact/signals": "1.1.1",
 		"ethers": "6.6.4",
+		"funtypes": "5.0.3",
 		"preact": "10.8.1"
 	},
 	"scripts": {
-		"build": "tsc",
+		"build": "rm -rf app/js && tsc",
 		"serve": "npx serve -L ./app",
 		"vendor": "npm install && npm --prefix ./build install && npm --prefix ./build run vendor",
```
KillariDev, line 18: instead of using "npm install" here, use "npm ci --ignore-scripts"

app/ts/schema.ts
jubalm: this is actually needed to stringify the data before sending to localStorage unlike StorageArea which accepts objects

package.json
jubalm: updated the lock file and package.json with this strategy

app/ts/schema.ts
0xjimmy:
```ts
let x: TokenListSchema
const stringifiedContent = x.safeSerialize() // The value should be string

**Review by 0xjimmy — APPROVED** [review]

File: app/ts/library/persistent-signal.ts [hunk 3]
```diff
@@ -0,0 +1,36 @@
...
+		if (cache) {
+			const parsed = funtypes.String.withParser(funTypeParser).safeParse(cache)
+			derivedSignal.value = parsed.success ? parsed.value : derivedSignal.peek()
+		}
+	}
+
+	const syncSignalToCache = () => {
+		const serializedStore = funtypes.String.withParser(funTypeParser).safeSerialize(derivedSignal.value)
+		if (!serializedStore.success) {
+			console.log('Failed to sync signal to cache')
```
KillariDev, line 21: this probably should be a warning print or a pure error?

app/ts/library/persistent-signal.ts
```diff
@@ -0,0 +1,36 @@
...
+			return
+		}
+
+		const stringCache = funtypes.String.safeParse(serializedStore.value)
+		if (!stringCache.success) {
+			console.log('serializedStore is not a string')
```
KillariDev, line 27: this probably should be a warning print or a pure error?

KillariDev (approved):

PR #197, opened 2023-08-29
Reimplement recent transfers to use localStorage

jubalm:
Requires branch `featrue/typescript-funtypes` to be merged

- [x] this changes the current implementation to  use funtypes to parse cache
- [x] ensure addresses are checksummed
- [x] remove unused codes (legacy implementation)


app/ts/schema.ts
```diff
@@ -23,21 +42,15 @@ export const TokenSchema = funtypes.Object({
...
-	},
-}
+export type Token = funtypes.Static<typeof TokenSchema>
 
-export const ManagedTokensCacheSchema = funtypes.String.withParser(ManagedTokensCacheParserConfig)
+export const TransferSchema = funtypes.Object({
+	hash: funtypes.String,
+	from: AddressSchema,
+	to: AddressSchema,
+	amount: funtypes.String,
```
KillariDev, line 51: this should be EthereumQuantity which you can get from interceptors wire types too

app/ts/schema.ts
```diff
@@ -23,21 +42,15 @@ export const TokenSchema = funtypes.Object({
...
-		if (!serializedValue.success) return { success: false, message: serializedValue.message }
-		const jsonString = JSON.stringify(serializedValue.value)
-		return { success: true, value: jsonString }
-	},
-}
+export type Token = funtypes.Static<typeof TokenSchema>
 
-export const ManagedTokensCacheSchema = funtypes.String.withParser(ManagedTokensCacheParserConfig)
+export const TransferSchema = funtypes.Object({
+	hash: funtypes.String,
```
KillariDev, line 48: we have `EthereumBytes32` in wiretypes of interceptor

app/ts/store/account.ts
```diff
@@ -4,20 +4,21 @@ import { ApplicationError } from './errors.js'
...
 	const provider = useProviders()
-	const { value: query, waitFor } = useAsyncState<string>()
+	const { value: query, waitFor } = useAsyncState<Address>()
 
 	const connect = () => {
 		waitFor(async () => {
 			try {
 				const signer = await provider.browserProvider.getSigner()
-				return getAddress(signer.address)
+				return AddressSchema.parse(getAddress(signer.address))
```
KillariDev, line 21: we could add helper function that does `AddressSchema.parse(getAddress())`

KillariDev (approved):

app/ts/schema.ts
jubalm: Creating a separate issue to tread addresses as bigint #200

app/ts/schema.ts
```diff
@@ -23,21 +42,15 @@ export const TokenSchema = funtypes.Object({
...
+	from: AddressSchema,
+	to: AddressSchema,
+	amount: funtypes.String,
+	token: funtypes.Union(TokenSchema, funtypes.Undefined),
+	date: funtypes.Number,
```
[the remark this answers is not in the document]
jubalm: I am not taking this from the block, just from the browser Date class to know when the transfer was initiated in Lunaria and not really on chain.

I may move it to read from chain so created an issue #201

app/ts/store/account.ts
jubalm: Created issue to add checksum validation for parser #202

PR #199, opened 2023-08-31
Reimplement favorites caching

jubalm:
- [ ] use persistent signal for caching


jubalm:
close in favor of #196

PR #209, opened 2023-10-06
remove account store dependency

jubalm:
remove old store and move account signal to context


app/ts/components/TransactionPage/index.tsx
```diff
@@ -1,22 +1,41 @@
...
+	const attemptToConnect = () => {
+		if (browserProvider === undefined) return
+		waitFor(async () => {
+			const [signer] = await browserProvider.listAccounts()
+			return EthereumAddress.parse(signer.address)
+		})
+	}
+
+	const listenForQueryChanges = () => {
+		// do not reset shared state for other instances of this hooks
```
KillariDev, line 32:
```suggestion
		// do not reset shared state for other instances of this hook
```

KillariDev (approved):

PR #214, opened 2023-10-17
Abbreviate huge numbers

jubalm:
Abbreviate huge numbers so that they don't clobber the UI.


app/ts/library/utilities.ts
```diff
@@ -64,3 +64,15 @@ export function areEqualStrings(a: string, b: string, caseSensitive?: true) {
 	if (caseSensitive) return a === b
 	return a.toLowerCase() === b.toLowerCase()
 }
+
+/**
+ * Abbreviate huge numbers
+ */
+export function abbreviateNumber(num: number): string {
+	const suffixes = ["", "K", "M", "B", "T", "Q"]
```
KillariDev, line 72:
```suggestion
	const suffixes = ['', 'k', 'M', 'G']
```
let's use SI units and not use the too big ones as they are not very familiar to people

app/ts/library/utilities.ts
```diff
@@ -64,3 +64,15 @@ export function areEqualStrings(a: string, b: string, caseSensitive?: true) {
 	if (caseSensitive) return a === b
 	return a.toLowerCase() === b.toLowerCase()
 }
+
+/**
+ * Abbreviate huge numbers
+ */
+export function abbreviateNumber(num: number): string {
```
KillariDev, line 71: we could abbreviate small numbers too

app/ts/library/utilities.ts
```diff
@@ -64,3 +64,15 @@ export function areEqualStrings(a: string, b: string, caseSensitive?: true) {
...
+	const suffixes = ["", "K", "M", "B", "T", "Q"]
+	const exponent = Math.floor(Math.log10(num) / 3)
+	const shortValue = num / Math.pow(1000, exponent)
+	const displayValue = shortValue.toFixed(1)
+	const suffix = suffixes[exponent]
+	return `${displayValue}${suffix}`
```
KillariDev, line 77: we need to handle the overflow here as well

KillariDev (changes requested):

jubalm:
- [ ] ensure tooltip for abbreviated numbers show entire number

app/ts/components/AbbreviatedValue.tsx
```diff
@@ -0,0 +1,16 @@
+import { parseNumericTerm } from '../library/utilities.js'
+
+export const AbbreviatedValue = ({ floatValue }: { floatValue: number }) => {
+	const { coefficient, exponent } = parseNumericTerm(floatValue)
+
+	if (exponent < 0) {
+		const leadingZeros = Math.abs(exponent) - 1
+		return <>0.<small>{'0'.repeat(leadingZeros)}</small>{coefficient}</>
```
KillariDev, line 8: if the input value (floatValue), is negative, this will show it as a positive number. In Interceptor we need to display negative values at times.

app/ts/library/utilities.ts
```diff
@@ -64,3 +64,33 @@ export function areEqualStrings(a: string, b: string, caseSensitive?: true) {
...
+		[1e9, 'G'],
+		[1e6, 'M'],
+		[1e3, 'k'],
+		[1, ''],
+	]);
+
+	for (const [factor, suffix] of siUnits) {
+		if (Math.abs(number) > factor) {
+			const shortValue = number / factor;
+			return `${shortValue.toFixed(precision)}${suffix}`;
```
KillariDev, line 82:
```suggestion
			const shortValue = number / factor
			return `${shortValue.toFixed(precision)}${suffix}`
```

app/ts/library/utilities.ts
```diff
@@ -64,3 +64,33 @@ export function areEqualStrings(a: string, b: string, caseSensitive?: true) {
...
+		}
+	}
+
+	return number.toString()
+}
+
+export function parseNumericTerm(num: number): { coefficient: number, exponent: number } {
+	const [c, e] = num.toExponential().split('e')
+	const exponent = parseInt(e)
+	return { coefficient: parseFloat(c), exponent };
```
KillariDev, line 92:
```suggestion
	return { coefficient: parseFloat(c), exponent }
```

app/ts/library/utilities.ts
```diff
@@ -64,3 +64,33 @@ export function areEqualStrings(a: string, b: string, caseSensitive?: true) {
...
+}
+
+const win = window as any
+win.parseNumericTerm = parseNumericTerm
```
KillariDev, line 96: what is this for?

KillariDev (changes requested):
fix the bug with small negative numbers (eg, -0.00001234). Otherwise looks good. I also do not know why do you need that window assigment

app/ts/components/AbbreviatedValue.tsx
jubalm: Created a separate issue here for negative numbers support #220

app/ts/library/utilities.ts
MicahZoltu: Also, never use `any`.

KillariDev (approved):

PR #219, opened 2023-11-07
Enhance error message UI

jubalm:
Allow user view more details of an error and copy raw error message which can be handy for debugging

| Screenshot 2023-11-07 at 9 12 53 AM |
|:--:|
| *default view* |

| Screenshot 2023-11-07 at 9 12 33 AM |
|:--:|
| *expanded view* |

### DIFF
--- a/app/ts/components/CopyButton.tsx
+++ b/app/ts/components/CopyButton.tsx
@@ -1,14 +1,19 @@
 import { useSignal, useSignalEffect } from '@preact/signals'
-import { ComponentChild } from 'preact'
+import { ComponentChild, JSX } from 'preact'
+import { useNotification } from '../context/Notification.js'
 import * as Icon from './Icon/index.js'
 
 type CopyButtonProps = {
 	value: string
-	label: ComponentChild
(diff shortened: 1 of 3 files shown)


app/ts/components/CopyButton.tsx
```diff
@@ -1,14 +1,19 @@
...
-import { ComponentChild } from 'preact'
+import { ComponentChild, JSX } from 'preact'
+import { useNotification } from '../context/Notification.js'
 import * as Icon from './Icon/index.js'
 
 type CopyButtonProps = {
 	value: string
-	label: ComponentChild
+	withLabel?: boolean
+	// for deprecation: use `withLabel` to display text label
```
KillariDev, line 9: add issue about this

KillariDev (approved):

app/ts/components/CopyButton.tsx
jubalm: created issue here: https://github.com/DarkFlorist/lunaria/issues/221

PR #225, opened 2023-11-16
IPFS Deployment

### LINKED ISSUE #218: Publish Lunaria to IPFS
Use the Dockerfile as a starting point
https://notes.ethereum.org/@GW1ZUbNKR5iRjjKYx6_dJQ/Bk8zsJ9xj#As-a-Developer

jubalm:
Resolves #218 

#### Notable changes

- [x] renamed the `build` folder to `vendor` so as to not confuse it as an output folder for the built application
- [x] the vendor package now exposes a command-line script `vendor` instead of running it directly using the `bin` property of `package.json`
- [x] the `build` script now includes the `vendor` command as this script injects the module dependencies for the app which is always necessary for it to work properly
- [x] regenerated package locks to reflect compatible version dependencies with `node@20.10` used in docker build

**Netlify was reconfigured for the changes in the build script*

### DIFF
--- a/.github/workflows/ipfs-deploy.yml
+++ b/.github/workflows/ipfs-deploy.yml
@@ -1,0 +1,18 @@
+name: Build and Push to IPFS
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v2
+
+      - name: Build and Push to IPFS
+        run: |
+          docker build -t lunaria:latest .
--- a/.github/workflows/ipfs-deploy.yml
+++ b/.github/workflows/ipfs-deploy.yml
@@ -2,17 +2,49 @@ name: Build and Push to IPFS
 
 on:
   push:
+    # Publish `main` as Docker `latest` image.
     branches:
-      - master
+      - main
+
+    # Publish `v1.2.3` tags as releases.
+    tags:
+      - v*
 
 jobs:
-  build:
+  # Push image to GitHub Packages.
+  # See also https://docs.docker.com/docker-hub/builds/
+  push:
     runs-on: ubuntu-latest
+    if: github.event_name == 'push'
 
     steps:
-      - name: Checkout Repository
-        uses: actions/checkout@v2
+      - uses: actions/checkout@v2
 
-      - name: Build and Push to IPFS
+      - name: Set IMAGE_TAG
+        run: echo "IMAGE_TAG=$(echo ${{ github.repository }} | tr '[A-Z]' '[a-z]')" >> $GITHUB_ENV
+
+      - name: Build image
         run: |
-          docker build -t lunaria:latest .
+          docker build . --file Dockerfile --tag $IMAGE_TAG
+
+      - name: Log into registry
+        run: echo "${{ secrets.CONTAINER_REGISTRY }}" | docker login ghcr.io -u ${{ github.actor }} --password-stdin
+
+      - name: Push image
+        run: |
+          IMAGE_ID=ghcr.io/$IMAGE_TAG
+
+          # Strip git ref prefix from version
+          VERSION=$(echo "${{ github.ref }}" | sed -e 's,.*/\(.*\),\1,')
+
+          # Strip "v" prefix from tag name
+          [[ "${{ github.ref }}" == "refs/tags/"* ]] && VERSION=$(echo $VERSION | sed -e 's/^v//')
+
+          # Use Docker `latest` tag convention
+          [ "$VERSION" == "main" ] && VERSION=latest
+
+          echo IMAGE_ID=$IMAGE_ID
+          echo VERSION=$VERSION
+
+          docker tag $IMAGE_TAG $IMAGE_ID:$VERSION
+          docker push $IMAGE_ID:$VERSION
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,7 @@ RUN npm ci
 COPY tsconfig.json tsconfig.json
 RUN npm run build
 
-# Create kubo container
+# Create kubo-based image
 FROM ipfs/kubo:latest as ipfs
 
 # Copy generated app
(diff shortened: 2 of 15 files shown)


Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM node:20.9.0-alpine as builder
+
+# Set directory for the sources to be processed
+WORKDIR /source
+
+# Install node depenedencies
+COPY app app
+COPY build build
```
jubalm, line 8: move this step down

Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+FROM ipfs/kubo:latest
+
+# Set path for lunaria static site
+WORKDIR /ipfs
+
+# Copy all built files from previous step
+COPY --from=builder /source/app .
+RUN ipfs init
+RUN ipfs add --cid-version 1 --quieter --only-hash --recursive . > ipfs_hash.txt
+RUN cat ipfs_hash.txt
```
jubalm, line 27: next step is to push it to a server

KillariDev (commented):
looks good for me, but Micah knows this stuff better so its better if he reviews :)

Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+# Create kubo container
+FROM ipfs/kubo:latest
+
+WORKDIR /container-init.d
+ADD ipfs.config.sh .
+RUN chmod a+x ipfs.config.sh
+
+# Copy generated app
+WORKDIR /var/www
+COPY --from=builder /source/app .
```
MicahZoltu, line 33: Why `/var/www`?  Seems like quite an arbitrary location.  The normal location for this in IPFS is `/export`.  Also, you don't need to set the `WORKDIR`, and doing so may break the running of the container.

```suggestion
COPY --from=builder /source/app /export
```

Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+
+RUN npm exec --prefix ./build vendor
+RUN npm run build
+
+# Create kubo container
+FROM ipfs/kubo:latest
+
+WORKDIR /container-init.d
+ADD ipfs.config.sh .
+RUN chmod a+x ipfs.config.sh
```
MicahZoltu, line 29: See comment below.  I recommend using the code I have in the Tornado Dockerimage as I don't think what you have here will do what we need.

.github/workflows/ipfs-deploy.yml
```diff
@@ -0,0 +1,18 @@
+name: Build and Push to IPFS
+
+on:
+  push:
+    branches:
+      - master
```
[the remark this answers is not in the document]
jubalm: updated the workflow yaml to build image with version tags

package.json
```diff
@@ -15,7 +15,7 @@
 	"scripts": {
 		"build": "rm -rf app/js && tsc",
 		"serve": "npx serve -L ./app",
-		"vendor": "npm ci --ignore-scripts && npm --prefix ./build ci --ignore-scripts && npm --prefix ./build run vendor",
-		"styles": "npm --prefix ./twcss ci --ignore-scripts && npm --prefix ./twcss run styles"
+		"vendor": "npm exec --prefix ./build vendor",
```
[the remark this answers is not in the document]
jubalm: i've combined the vendor script with build script since it's necessary to run the app anyways and `npm install` (or `npm ci`) should have been ran by then as well

ipfs.config.sh
```diff
@@ -0,0 +1,3 @@
+ipfs config Addresses.API --json '["/ip4/0.0.0.0/tcp/5001", "/ip6/::/tcp/5001"]'
+ipfs config --json Addresses.AppendAnnounce '["/ip4/'"${IPV4_ADDRESS:-0.0.0.0}"'/tcp/4001"]'
+ipfs add --cid-version 1 --recursive /var/www
```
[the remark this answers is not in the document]
jubalm: I was aiming to build an image that could be ran/hosted and will automatically host Lunaria given an IPV4 address but bailed. this should now be removed in favor of exposing the base ipfs/kubo image, and only adding the app to `/export` folder

Dockerfile
jubalm: changed copy path to `/export` as per suggestion on a separate commit

Dockerfile
jubalm: this script/file should not be on the latest changes to this PR

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+RUN npm ci
+
+# Copy project files and run necessary symlinks npm needs
+COPY app app
+COPY vendor vendor
+RUN npm ci
+
+# Buld the app
+COPY tsconfig.json tsconfig.json
+RUN npm run build
```
MicahZoltu, line 22: A few changes in here:
1. Path basing can be confusing, so I recommend being quite explicit.  `./` for "root of Docker build context" and `/app/` for the target directory in the container as an absolute path.  You could do `/source/` for the target directory in the container (or anything else really), but `/app` I think is a fairly common choice for "where in the container do you put your app" so people may look there first.  Another common one is `workspace` I think.
2. I removed the excess `npm ci` because it shouldn't be necessary.  Let me know if it is needed for some reason.
3. Add running of the vendor script, which appears to be missing.
4. Be explicit about which folders are copied.  This is important because if you just copy the whole `./app` folder then you'll *also* copy over the `js` files.  Our build script deletes them, but one can imagine other folders with generated data in them later, and being explicit ensures we don't accidentally copy anything we shouldn't be.

```suggestion
# Install vendor dependencies
COPY ./vendor/package.json /app/vendor/package.json
COPY ./vendor/package-lock.json /app/vendor/package-lock.json
WORKDIR /app/vendor
RUN npm ci

# Install app dependencies
COPY ./package.json /app/package.json
COPY ./package-lock.json /app/package-lock.json
WORKDIR /app
RUN npm ci

# run vendor script
COPY ./vendor/ /app/vendor/
WORKDIR /app/build
RUN npm run vendor

# Buld the app
COPY ./tsconfig.json /app/tsconfig.json
COPY ./index.html /app/app/index.html
COPY ./app/css/ /app/app/css/
COPY ./app/img/ /app/app/img/
COPY ./app/ts/ /app/app/ts/
WORKDIR /app
RUN npm run build
```

vendor/index.js
MicahZoltu, line 1: We shouldn't be committing JS files to the repository, just TS.

PR #229, opened 2023-11-23
Fix balance rendering issue

### LINKED ISSUE #228: ETH balance not displayed correctly.
Use interceptor to look at Vitalik's account and notice that Lunaria only shows him as having 7.842 ETH, 7.029 USDC, and 3.112 WETH.  In reality, he has 784.2 ETH, 70.29 USDC, and 31.12 WETH.  It appears that numbers are always being shifted to 1 decimal digit visible.

Also, WBTC is showing with two decimal places, which may be a related bug.

image

jubalm:
Resolves #228 

Screenshot 2023-11-24 at 3 55 30 AM

### DIFF
--- a/app/ts/components/AbbreviatedValue.tsx
+++ b/app/ts/components/AbbreviatedValue.tsx
@@ -23,6 +23,6 @@ export const AbbreviatedValue = ({ floatValue }: { floatValue: number }) => {
 }
 
 function toFixedLengthDigits(num: number, max: number = 5) {
-	const formatter = new Intl.NumberFormat('en-US', { maximumSignificantDigits: max })
+	const formatter = new Intl.NumberFormat('en-US', { maximumSignificantDigits: max, useGrouping: false })
 	return formatter.format(num)
 }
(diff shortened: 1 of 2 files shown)


app/ts/components/AbbreviatedValue.tsx
```diff
@@ -1,16 +1,28 @@
...
+		if (floatValue >= prefix.value) {
+			return <>{toFixedLengthDigits(floatValue / prefix.value) + prefix.symbol}</>
+		}
+	}
 
-	if (exponent < 0) {
-		const leadingZeros = Math.abs(exponent) - 1
-		return <>0.<small>{'0'.repeat(leadingZeros)}</small>{coefficient}</>
+	// if value is a fraction of 1
+	if (floatValue && floatValue % 1 === floatValue) {
```
KillariDev, line 15: if floatValue = 100.1, this prints wrong values?

I feel we should have test cases for this calculation and it should be separated from the preact code so we can run tests for it

KillariDev (changes requested):

app/ts/components/AbbreviatedValue.tsx
jubalm: this condition are for floats between 0 and 1 such as 0.0001; takes care of rendering the zeros after the decimal point as smaller font size.

KillariDev (approved):

PR #248, opened 2024-01-11
Added support for hosting from a subfolder

### LINKED ISSUE #246: Image paths don't work when hosted as `/ipfs`

jubalm:
- dynamically change the HTML base URL to the window's location
- changed absolute `/` paths to relative path `./`

Resolves #246

There may still be errors in the page before the base href was set but most modern browsers will find the files after the switch; at least on Chrome and Firefox was tested ok

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -8,9 +8,12 @@
 
 	<!-- support hosting from a subpath such as ipfs -->
 	<script>
-		const baseElement = document.createElement('base')
-		baseElement.setAttribute('href', window.location.href.replace(/\/+$/, '') + '/')
-		document.head.appendChild(baseElement)
+		const subPath = window.location.pathname.substring(0, window.location.pathname.lastIndexOf('/'))
+		if (subPath) {
+			const baseElement = document.createElement('base')
+			baseElement.setAttribute('href', `${subPath}/`)
+			document.head.appendChild(baseElement)
+		}
 	</script>
 
 	<link rel="icon" type="image/svg+xml" href="./img/icon-lunaria.svg" />
--- a/app/index.html
+++ b/app/index.html
@@ -8,7 +8,7 @@
 
 	<!-- support hosting from a subpath such as ipfs -->
 	<script>
-		const subPath = window.location.pathname.substring(0, window.location.pathname.lastIndexOf('/'))
+		const subPath = window.location.pathname.replace('index.html', '').replace(/\/$/, '')
 		if (subPath) {
 			const baseElement = document.createElement('base')
 			baseElement.setAttribute('href', `${subPath}/`)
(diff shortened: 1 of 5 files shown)


MicahZoltu (approved):
Approved, but I would like more reasoning/discussion on the added script as it feels unnecessary.

app/index.html
```diff
@@ -5,6 +5,14 @@
 	<title>Lunaria</title>
 	<meta charset='utf-8' />
 	<meta name='viewport' content='width=device-width, initial-scale=1.0' />
+
+	<!-- support hosting from a subpath such as ipfs -->
+	<script>
+		const baseElement = document.createElement('base')
+		baseElement.setAttribute('href', window.location.href.replace(/\/+$/, '') + '/')
+		document.head.appendChild(baseElement)
+	</script>
```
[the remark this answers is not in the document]
jubalm: Replaced `location.href` strategy of parsing subpath by using `location.pathname`

```js
window.location.pathname.substring(0, window.location.pathname.lastIndexOf('/'))
```

[This commit](https://github.com/DarkFlorist/lunaria/pull/248/commits/12f48eeb838e33b0ff733c71c5862823139b3738) enable support for hosting as `/subpath/`, `/subpath/`, `/subpath/index.html` and appropriately injects only when needed.

app/index.html
```diff
@@ -5,6 +5,17 @@
 	<title>Lunaria</title>
 	<meta charset='utf-8' />
 	<meta name='viewport' content='width=device-width, initial-scale=1.0' />
+
+	<!-- support hosting from a subpath such as ipfs -->
+	<script>
+		const subPath = window.location.pathname.substring(0, window.location.pathname.lastIndexOf('/'))
```
MicahZoltu, line 11:
```ts
'/subfolder'.lastIndexOf('/') === 0
```
I don't think this will do what you want as it will strip the `subfolder`, but we want to keep it.

MicahZoltu:
```suggestion
		const subPath = window.location.pathname.endsWith('index.html')
			? window.location.pathname.substring(0, window.location.pathname.length - 'index.html'.length)
			: window.location.pathname
```
I think something like this may work?  Double check that a base href of '/subfolder' and '/subfolder/' work the same though, as this will not add a trailing slash if it is missing.

jubalm: I just made it so that it will always remove `index.html` if it's in the path and just take the subpath from out of it by removing the trailing slash. will re-add it later as needed by the base element.

```ts
const subPath = window.location.pathname.replace('index.html', '').replace(/\/$/, '')
```

Screenshot 2024-01-18 at 12 50 40 PM

MicahZoltu (approved):
Looks good!

PR #249, opened 2024-01-11
Warn users hosting Lunaria from a subpath

### LINKED ISSUE #245: Redirect lunaria from subpath to subdomain resolution

jubalm:
Screenshot 2024-01-11 at 1 56 19 PM

Resolved #245

### DIFF
--- a/app/ts/components/SubpathHostWarning.tsx
+++ b/app/ts/components/SubpathHostWarning.tsx
@@ -1,0 +1,35 @@
+import { useSignalEffect } from '@preact/signals'
+import { useSignalRef } from '../library/preact-utilities.js'
+
+export const SubpathHostWarning = () => {
+	const { ref, signal: dialogRef } = useSignalRef<HTMLDialogElement | null>(null)
+
+	useSignalEffect(() => {
+		if (!dialogRef.value) return
+
+		// hosting on root is ok
+		if (window.location.pathname === '/') return
+
+		// if pathname is other than the root host, show warning
+		dialogRef.value.showModal()
+	})
+
+	return (
+		<dialog ref={ref} class='w-full text-white backdrop:bg-black/80 backdrop:backdrop-blur-[2px] max-w-full max-h-full md:max-w-lg md:max-h-[calc(100vh-3rem)] bg-transparent'>
+			<form method='dialog' class='px-6'>
+				<div class='text-lg font-bold text-center mb-4'>Warning! Insecure Host Detected.</div>
+				<div class='text-center'>
+					<p class='mb-4'>Hosting Lunaria from a subpath <pre class='inline px-2 py-[3px] bg-white/20 text-sm'>{window.location.pathname}</pre> highly discouraged as this site is uses browser storage. By doing so, your data may be accessed by the host domain.</p>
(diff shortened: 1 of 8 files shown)


app/ts/components/SubpathHostWarning.tsx
```diff
@@ -0,0 +1,35 @@
...
+		dialogRef.value.showModal()
+	})
+
+	return (
+		<dialog ref={ref} class='w-full text-white backdrop:bg-black/80 backdrop:backdrop-blur-[2px] max-w-full max-h-full md:max-w-lg md:max-h-[calc(100vh-3rem)] bg-transparent'>
+			<form method='dialog' class='px-6'>
+				<div class='text-lg font-bold text-center mb-4'>Warning! Insecure Host Detected.</div>
+				<div class='text-center'>
+					<p class='mb-4'>Hosting Lunaria from a subpath <pre class='inline px-2 py-[3px] bg-white/20 text-sm'>{window.location.pathname}</pre> highly discouraged as this site is uses browser storage. By doing so, your data may be accessed by the host domain.</p>
+					<p class='mb-4'>If hosting this through IPFS, consider access using subdomain resolution which is the <a class='outline-none underline underline-offset-2' href='https://docs.ipfs.tech/how-to/gateway-best-practices/#avoiding-centralization'>recommended approach</a>, according to it&apos;s documentation.</p>
```
MicahZoltu, line 23:
```suggestion
					<p class='mb-4'>Hosting Lunaria from a subpath <pre class='inline px-2 py-[3px] bg-white/20 text-sm'>{window.location.pathname}</pre> highly discouraged as this site is uses browser storage. By doing so, your data may be accessed by other servers on the same domain.</p>
					<p class='mb-4'>If hosting this through IPFS, consider access using subdomain resolution instead which is the <a class='outline-none underline underline-offset-2' href='https://docs.ipfs.tech/how-to/gateway-best-practices/#avoiding-centralization'>recommended approach</a>.</p>
```

KillariDev: could we link to the right way of doing the hosting instead, by swapping the urls? so user could click it to get the right way?

KillariDev (approved):

app/ts/components/SubpathHostWarning.tsx
MicahZoltu: We would need to get the CID converter logic in our app, which I believe is non-trivial.  If we had that, we could detect an IPFS CIDv0 in the path, convert it to a CIDv1, and generate a link to the assumed subdomain routed version of the same page.  It appears this library can do the conversion for us, and has no transitive dependencies so I wouldn't be opposed to using it: https://www.npmjs.com/package/multiformats

MicahZoltu: Decoding algorithm if we want to do it by hand: https://github.com/multiformats/cid/blob/ef1b2002394b15b1e6c26c30545fd485f2c4c138/README.md#decoding-algorithm

Converting using multiformats library:
```ts
const v0 = CID.parse('QmdfTbBqBPQ7VNxZEYEj14VmRuZBkqFbiwReogJgS1zR1n')
v0.toString()
//> 'QmdfTbBqBPQ7VNxZEYEj14VmRuZBkqFbiwReogJgS1zR1n'
v0.toV1().toString()
//> 'bafybeihdwdcefgh4dqkjv67uzcmw7ojee6xedzdetojuzjevtenxquvyku'
```

jubalm:
Closing in favor of redirecting the user to subdomain resolution

PR #251, opened 2024-01-18
Redirect IPFS subpath using subdomain resolution

jubalm:
UPDATED:
This should now automatically redirect when an IPFS subpath is detected instead of showing a warning

### DIFF
--- a/app/ts/components/IPFSSubpathRedirect.tsx
+++ b/app/ts/components/IPFSSubpathRedirect.tsx
@@ -1,0 +1,34 @@
+import { CID } from 'multiformats'
+import { useEffect } from 'preact/hooks'
+
+export const IPFSSubpathRedirect = () => {
+	function extractHashFromBasePath() {
+		const htmlBase = document.querySelector('base')
+		if (!htmlBase || !htmlBase.href.includes('/ipfs/')) return
+		const cidRegex = /\/ipfs\/(Qm[1-9A-HJ-NP-Za-km-z]{44,}|b[A-Za-z2-7]{58,}|B[A-Z2-7]{58,}|z[1-9A-HJ-NP-Za-km-z]{48,}|F[0-9A-F]{50,})/
+		const [_, maybeCidMatch] = htmlBase.href.match(cidRegex) || []
+		return maybeCidMatch
+	}
+
+	function generateIPFSRedirectUrl() {
+		const cidFromPath = extractHashFromBasePath()
+		console.log(cidFromPath)
+
+		if (!cidFromPath) {
+			console.log('No IPFS subpath detected')
+			return
+		}
+
+		const v1CidString = CID.parse(cidFromPath).toV1().toString()
+		if (!v1CidString) return
+		return `${location.protocol}//${v1CidString}.ipfs.${location.host}`
+	}
+
+	useEffect(() => {
+		const redirectUrl = generateIPFSRedirectUrl()
+		if (!redirectUrl) return
+		window.location.href = redirectUrl
+	}, [])
+
+	return <></>
+}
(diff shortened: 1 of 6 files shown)


KillariDev (approved):

app/ts/components/IPFSSubpathRedirect.tsx
```diff
@@ -0,0 +1,31 @@
+import { CID } from 'multiformats'
+import { useEffect } from 'preact/hooks'
+
+export const IPFSSubpathRedirect = () => {
+	function extractHashFromBasePath() {
+		const htmlBase = document.querySelector('base')
+		if(!htmlBase) return
+		const cidRegex = /\/ipfs\/(Qm[1-9A-HJ-NP-Za-km-z]{44,}|b[A-Za-z2-7]{58,}|B[A-Z2-7]{58,}|z[1-9A-HJ-NP-Za-km-z]{48,}|F[0-9A-F]{50,})/
```
MicahZoltu, line 8: Did you come up with this, or did you find it online somewhere?

MicahZoltu: Is there no function in multiformats like `isCid`?

jubalm: > Did you come up with this, or did you find it online somewhere?

https://stackoverflow.com/questions/67176725/a-regex-json-schema-pattern-for-an-ipfs-cid
it was linked from some writing/blog but couldn't trace back which one i got it from 😖

MicahZoltu: Consider using https://multiformats.github.io/js-multiformats/classes/cid.CID.html#asCID-2 instead.  According to docs, it will return null if the input value is not a CID.

Something like this (pseudocode, untested):
```ts
if !(path.contains('/ipfs/')) return
const maybeCidMatch = /\/ipfs\/(.*?)\//.exec(path)
if (!maybeCidMatch) return
const cid = CID.asCID(maybeCidMatch[1]!)
return cid
```

jubalm: This part of the code only extracts the hash/cid from base element that should be parsed by another function so simplified the regex match

```ts
const htmlBase = document.querySelector('base')
if (!htmlBase || !htmlBase.href.includes('/ipfs/')) return
const maybeCidMatch = /\/ipfs\/(.*?)\//.exec(htmlBase.href)
return maybeCidMatch?.at(1)
```

app/ts/components/IPFSSubpathRedirect.tsx
```diff
@@ -0,0 +1,36 @@
...
+			return
+		}
+
+		try {
+			const v1CidString = CID.parse(cidFromPath).toV1().toString()
+			return `${location.protocol}//${v1CidString}.ipfs.${location.host}`
+		} catch (error) {
+			console.log('Failed to transform string to CID V1', `${error}`)
+			return
+		}
```
MicahZoltu, line 26: We shouldn't be using `try/catch` for code flow, and if the path segment following `/ipfs/` isn't a CID then we shouldn't try to redirect.
```suggestion
		const cid = CID.asCID(cidFromPath)
		if (!cid) return
		const v1CidString = cid.toV1().toString()
		return `${location.protocol}//${v1CidString}.ipfs.${location.host}`
```
Double check this code works as the docs say and returns `undefined` in the case where the path segment is not a CID, and returns a correct URL when it is either a CIDv0 or CIDv1, but this lets us avoid using a try/catch.

jubalm: the `asCID` function does not work with CID as strings, only with CID objects generated by `CID.parse` or similar libraries

MicahZoltu: This makes me sad, but we can move back to the regex then.  We still shouldn't be using `try/catch` for control flow, so before we call `CID.parse(...)` we should already be very confident we have an IPFS content hash in `cidFromPath`, which means using that regex you had previously.

PR #254, opened 2024-02-08
IPFS host using nft.storage

jubalm:
The script would need an nft.storage token to run

- [x] use nft.storage javascript api to upload
- [x] change strategy to use HTTP api endpoints instead of JS library
- [x] incorporate with a github workflow


ipfs-upload/upload.ts
```diff
@@ -0,0 +1,60 @@
+import { NFTStorage, File } from 'nft.storage';
```
MicahZoltu, line 1: We need to figure out how to do this without using this library, as it has over 200 transitive dependencies.

.github/workflows/ipfs-deploy.yml
```diff
@@ -29,7 +29,7 @@ jobs:
 
       - name: Build image
         run: |
-          docker build . --file Dockerfile --tag $IMAGE_TAG
+          docker build --target self-host . --file Dockerfile --tag $IMAGE_TAG
```
MicahZoltu, line 32: We need a job step (probably the last one) for `docker ebuild --target remote-host`.

Dockerfile
```diff
@@ -44,3 +48,21 @@ RUN printf '#!/bin/sh\nipfs --api /ip4/`getent ahostsv4 host.docker.internal | g
...
+# add the build output to IPFS, write the hash to a file and print during build
+RUN ipfs add --cid-version 1 --quieter --recursive /export > ipfs_hash.txt && cat ipfs_hash.txt
+
+# nft.storage api key required, pass the key during runtime with -e flag
+ENV NFTSTORAGE_API_KEY=
+
+# generate the car file and upload to nft.storage through http endpoint
+RUN printf '#!/bin/sh\nif [ -z "$NFTSTORAGE_API_KEY" ] || [ "$NFTSTORAGE_API_KEY" = "" ]; then echo "Error: NFTSTORAGE_API_KEY is not set or is blank"; exit  1; fi\nipfs dag export $(cat ipfs_hash.txt) > output.car\ncurl -X POST -H "Authorization: Bearer $NFTSTORAGE_API_KEY" -H "Content-Type: application/car" --data-binary @output.car https://api.nft.storage/upload' >> entrypoint.sh && chmod u+x entrypoint.sh
+
+ENTRYPOINT [ "./entrypoint.sh" ]
```
MicahZoltu, line 68: We should consider moving this into separate RUN lines and not having an `ENTRYPOINT`.  This would cause the docker image build process to submit to nft.storage.  We would then discard the image after that.  Essentially it just makes it so the image build process is effectively executing some code inside a container.  The nice thing about this strategy is that it keeps all of the build scripts inside a `Dockerimage` file and easy to read rather than spread out.

MicahZoltu: Another option (we can discuss merits on Discord) would be to have a single image and a more complex script that lets you do *either* local deploy or remote deploy.  We would publish the image and then the last CI step would be to run the built image with the parameter to trigger remote deploy.

This second option is growing on me.

Dockerfile
```diff
@@ -19,28 +23,96 @@ COPY ./app/img/ /source/app/img/
...
+      echo "NFTSTORAGE_API_KEY environment variable is not set";
+      exit  1;
+    fi
+
+    # Show the IFPS build hash
+    echo "Build Hash: $(cat /ipfs_hash.txt)"
+
+    # Create a CAR archive from build hash
+    echo "Uploading files to nft.storage..."
+		IPFS_HASH=$(ipfs add --cid-version 1 --quieter -r /export)
```
MicahZoltu, line 86: Mixed tabs and spaces.  Repository is otherwise all tab indents, so we should do tab indents in this file as well.

Dockerfile
```diff
@@ -19,28 +23,96 @@ COPY ./app/img/ /source/app/img/
...
+    echo "Invalid option: $1"
+    echo "Example usage: docker run --rm ghcr.io/darkflorist/lunaria:latest [docker-host|nft.storage]"
+    exit  1
+    ;;
+esac
+EOF
+
+RUN chmod u+x /entrypoint.sh
 
 ENTRYPOINT [ "./entrypoint.sh" ]
```
MicahZoltu, line 117: These two lines need to reference the same file.  One refers to `/entrypoint.sh` and the other refers to `./entrypoint.sh`.  It appears the file is written to `/entrypoint.sh`, so this is probably what you want:
```suggestion
RUN chmod u+x /entrypoint.sh

ENTRYPOINT [ "/entrypoint.sh" ]
```

MicahZoltu: We should also specify the required environment variable of `NFTSTORAGE_API_KEY` with:
```
ENV NFTSTORAGE_API_KEY
```
This will make it so tools will prompt the user for the variable rather than the user needing to guess it is necessary.

jubalm:
Tested locally with IPFS Desktop and Remotely with a private API key

***Can ignore the platform warning as this was ran on silicon chip mac*

| Screenshot 2024-02-15 at 2 14 31 PM |
|:--:|
|**docker upload**|

| Screenshot 2024-02-15 at 2 16 24 PM |
|:--:|
|**nft.storage upload**|

Dockerfile
```diff
@@ -19,28 +23,96 @@ COPY ./app/img/ /source/app/img/
...
+			-s \
+			https://api.nft.storage/upload)
+
+		# Show link to nft.storage (https://xxx.ipfs.nftstorage.link)
+		UPLOAD_SUCCESS=$(echo "$UPLOAD_RESPONSE" | jq -r ".ok")
+
+		if [ "$UPLOAD_SUCCESS" = "true" ]; then
+			echo "Succesfully uploaded to https://"$(echo "$UPLOAD_RESPONSE" | jq -r ".value.cid")".ipfs.nftstorage.link"
+		else
+			echo "Upload Failed: " $(echo "$UPLOAD_RESPONSE" | jq -r ".error.message")
```
MicahZoltu, line 103: Recommend printing the entire response here after the `error.message`.  The response may contain other useful information for debugging.

Dockerfile
```diff
@@ -19,28 +23,96 @@ COPY ./app/img/ /source/app/img/
...
+		IPFS_HASH=$(ipfs add --cid-version 1 --quieter -r /export)
+    ipfs dag export $IPFS_HASH > output.car
+
+    # Upload the entire directory to nft.storage
+		UPLOAD_RESPONSE=$(curl \
+			-X POST -H "Authorization: Bearer $NFTSTORAGE_API_KEY" \
+			-H "Content-Type: application/car" \
+			--data-binary @output.car \
+			-s \
+			https://api.nft.storage/upload)
```
MicahZoltu, line 95: Recommend using verbose parameters for enhanced clarity.
```suggestion
		UPLOAD_RESPONSE=$(curl \
			---request POST \
			--header "Authorization: Bearer $NFTSTORAGE_API_KEY" \
			--header "Content-Type: application/car" \
			--data-binary @output.car \
			--silent \
			https://api.nft.storage/upload)
```

Dockerfile
```diff
@@ -19,28 +23,96 @@ COPY ./app/img/ /source/app/img/
...
 RUN ipfs init
+RUN ipfs add --cid-version 1 --quieter --only-hash -r /export > ipfs_hash.txt
 
-# Copy lunaria build output
-COPY --from=builder /source/app /export
+# --------------------------------------------------------
+# Publish Script: Option to host app locally or on nft.storage
+# --------------------------------------------------------
+
+COPY <<'EOF' /entrypoint.sh
```
MicahZoltu, line 52: Recommend setting the working directory since this script will create some files along the way and it would be good to be explicit about where they end up.  I think `~` will work, but you could set it to literally any directory including one you make up like `/workspace` which is a common choice.
```suggestion
WORKDIR ~
COPY <<'EOF' /entrypoint.sh
```

Dockerfile
```diff
@@ -19,28 +23,96 @@ COPY ./app/img/ /source/app/img/
...
+# --------------------------------------------------------
+# Base Image: Create the base image that will host the app
+# --------------------------------------------------------
+
 # Cache the kubo image
 FROM ipfs/kubo:v0.25.0@sha256:0c17b91cab8ada485f253e204236b712d0965f3d463cb5b60639ddd2291e7c52 as ipfs-kubo
 
 # Create the base image
-FROM debian:12.2-slim@sha256:93ff361288a7c365614a5791efa3633ce4224542afb6b53a1790330a8e52fc7d
+FROM debian:12.2-slim@sha256:93ff361288a7c365614a5791efa3633ce4224542afb6b53a1790330a8e52fc7d as base
```
MicahZoltu, line 34: No need for the named image anymore.
```suggestion
FROM debian:12.2-slim@sha256:93ff361288a7c365614a5791efa3633ce4224542afb6b53a1790330a8e52fc7d
```

MicahZoltu (approved):
Overall looks good.  Left a handful of minor points of feedback that should be addressed or responded to before merging but overall it is ready.

.github/workflows/ipfs-deploy.yml
jubalm: this should now be reverted in favor of adding parameters to `docker run` strategy

.github/workflows/ipfs-deploy.yml
```diff
@@ -57,3 +57,11 @@ jobs:
 
           docker tag $IMAGE_TAG $IMAGE_ID:$VERSION
           docker push $IMAGE_ID:$VERSION
+
+					# Make image reference available to subsequent steps
+					echo "IMAGE_RELEASE_ID=$(echo $IMAGE_ID:$VERSION)" >> $GITHUB_ENV
```
MicahZoltu, line 62: Indentation is broken again here.  You may want to verify you have EditorConfig extension for your IDE, as that should use tabs everywhere except YAML, which should be spaces (because YAML is dumb).

MicahZoltu: Oops, looks like editorconfig wasn't setup properly.  https://github.com/DarkFlorist/lunaria/pull/256

Dockerfile
jubalm: changed upload error message to return full error json object `$(echo "$UPLOAD_RESPONSE" | jq -r ".error | @json")`

PR #267, opened 2024-04-23
Updated abbreviate values component

jubalm:
Sync this component with The Interceptor implementation


KillariDev:
I guess we should wait until we get it merged to interceptor

KillariDev (approved):

PR #278, opened 2024-08-08
Allow erc20 token address in full or lower letter case

jubalm:
Screenshot 2024-08-08 at 3 21 52 PM

### DIFF
--- a/app/ts/components/TokenAdd.tsx
+++ b/app/ts/components/TokenAdd.tsx
@@ -182,6 +182,8 @@ const QueryAddressField = () => {
 const QueryStatus = () => {
 	const { address, tokenQuery } = useTokenQuery()
 
+	const noHexPrefix = (hexAddress: string) => hexAddress.replace(/^0x/, '')
+
 	switch (tokenQuery.value.state) {
 		case 'inactive':
 			return <></>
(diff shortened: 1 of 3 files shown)


KillariDev (approved):
image

This allows you to add a token with a wrong checksum thought, as it seems like checksum check is not done

jubalm:
> image
> This allows you to add a token with a wrong checksum thought, as it seems like checksum check is not done

This should be fixed with [7feda45](https://github.com/DarkFlorist/lunaria/pull/278/commits/7feda457a65305a731ebb97e5ab9b0d958bca95e)

app/ts/components/TokenAdd.tsx
```diff
@@ -205,16 +207,24 @@ const QueryStatus = () => {
 				</div>
 			)
 		case 'resolved':
-			const token = serialize(ERC20Token, tokenQuery.value.value)
+			// queried token will have included a checksum address
+			const fetchedToken = tokenQuery.value.value
+			const showChecksumWarning = useComputed(() => {
+				const userInputAddress = noHexPrefix(address.value)
```
KillariDev, line 213: why do you remove the hex prefix? this also fails to remove `0X`

KillariDev (approved):

app/ts/components/TokenAdd.tsx
jubalm: it's now using ethersjs checksum validation which should already do this actually

jubalm:
The fields should now correctly check for address checksum and accepts all lower-case or upper-case addresses

| Screenshot 2024-09-20 at 8 44 34 PM | image |
|--|--|
| *token add* | *send* |

KillariDev (approved):

